PR #9, opened 2021-07-24
Update channels.py

bradleysigma:


JamesDearlove (commented):
Unfortunately, no this does not solve the issue. You'll need something like this: https://discordpy.readthedocs.io/en/stable/ext/commands/commands.html#variable

bradleysigma:
Rerequested with #18.

PR #19, opened 2021-09-11
Create Gaming COG - !bgg and !scry commands

bradleysigma:
image

image

PR #20, opened 2021-09-11
Create Text COG - !binify, !caesar, !httpcat, !mock and !zalgo commands 

bradleysigma:
image

image

image

image

image


JamesDearlove (changes requested):
Looks good. Only nit is as PR #19, the description for a command only needs what the command does, not command syntax.

JamesDearlove (approved):

PR #21, opened 2021-09-11
Autowave

banguyen10111999:
image
image

Wave react on member join event only on new member system message.


banguyen10111999:
image

image

Was done with channel name configured to "alpha", changed to "intros" on commit.

JamesDearlove (approved):
Nice work, looks all good to me. Will merge in pronto

PR #24, opened 2021-10-23
Change !joinchannel to support multiple channels.

kentonlam:
This makes !joinchannels redundant but keeps it as a hidden alias just in case.


JamesDearlove (approved):
lgtm

PR #26, opened 2021-10-28
Allow pings for working_on

TomCranitch:


JamesDearlove (approved):
Added a change in for jobs-bulletin as well

PR #27, opened 2021-11-04
Attempt at fixing working_on pings

andrewj-brown:
I suspect the "Users" ping perm is defaulting to the general setting of none; this *might* fix the problem of working-on pings going to random, unknown users


TomCranitch:
turns out I should have read the docs better when I made the permissions changes. for future reference the allowed_mentions object is merged with the one thats used to initalised it (as opposed to overriding it) as described [here](https://discordpy.readthedocs.io/en/stable/api.html?highlight=allowedmentions#discord.abc.Messageable.send) 

image

that being said im still a bit confused as to why it works on the testing server. i think it might be something weird happening with the larger member list on the main server. either way the bot should be allowed to ping people so we should merge this and i guess we can hope it fixes things as well

TomCranitch (approved):

PR #28, opened 2021-11-04
Use the channel's members instead of the bot's members for WorkingOn

PureGero:
The bot is pinging unknown users for the working-on, this may be because it's accessing its own list of members instead of just the members within the general channel. This should fix it by using the general channel's member list instead.


JamesDearlove:
I believe going forward the general channel will be a non-removable channel, so this won't (and isn't currently) an issue.

Thanks for your PR though!

PR #29, opened 2021-11-18
allowed !joinchannel to asccept names leading with hashes

antdon:
I'm not really sure how to test whether my code works. Hopefully this removes the the hash before a channel name such that it can be dealt with normally. In response to #29


JamesDearlove (approved):
lgtm

PR #34, opened 2021-11-30
Add a basic error handler

TomCranitch:
Sends a message if a required positional argument is missing. All errors also get logged


JamesDearlove (approved):
Great idea, lgtm.

PR #37, opened 2022-01-18
Added scary text

Thomas-Malcolm:
"Added" "scary" "text"

### DIFF
--- a/uqcsbot/text.py
+++ b/uqcsbot/text.py
@@ -81,6 +81,14 @@ class Text(commands.Cog):
         await ctx.send("".join(choice((c.upper(), c.lower())) for c in " ".join(text)))
 
     @commands.command()
+    async def scare(self, ctx: commands.Context, *text: str):
+        """
+        "adds" "scary" "quotes" "around" "the" "provided" "text"
+        """
+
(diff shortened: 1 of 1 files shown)


uqcsbot/text.py
```diff
@@ -80,6 +80,14 @@ async def mock(self, ctx: commands.Context, *text: str):
         """
         await ctx.send("".join(choice((c.upper(), c.lower())) for c in " ".join(text)))
 
+    @commands.command()
+    async def scare(self, ctx: commands.Context, *text: str):
+        """
+        "adds" "scary" "quotes" "around" "the" "provided" "text"
+        """
+
+        await ctx.send(" ".join(f'"{w}"' for w in text.split(' ')))
```
TomCranitch, line 89:
```suggestion
        await ctx.send(" ".join(f'"{w}"' for w in text))
```

TomCranitch (changes requested):

uqcsbot/text.py
kentonlam: won't this just put quotes around every character?

kentonlam: nvm

TomCranitch: can we make !extrascare that does that pls

PR #39, opened 2022-03-04
Azure Migration with Web Server

JamesDearlove:
This PR redirects the deployment to an Azure App Service. With this, a web server has been added so that Azure has something to confirm the bot remains running.


TomCranitch (approved):
haven't checked the gh workflow but otherwise lgtm

PR #43, opened 2022-03-07
React to Join Channels

campbell-mcfadden:
Creates a message that you can react to in order to join/leave channels.

``!joinmessage`` creates the message listing all channels available to join and their corresponding emojis. The bot then reacts to the message with each channel emoji. Only users with manage_channel perms can use this command.

Upon reacting to the message the user gets added/removed from $channel, depending on if they're already in it or not. Reactions are automatically removed by the bot so the counters stay at 1 (so as to maintain privacy, stop reacts from reaching absurd numbers etc). A dm confirming they joined/left is sent as well.

This doesn't affect the other commands for managing channels so they can still be used.


uqcsbot/join.py
```diff
@@ -0,0 +1,103 @@
...
+
+from uqcsbot.bot import UQCSBot
+from uqcsbot.models import Channel
+
+JOINED_PERMISSIONS = discord.Permissions(read_messages=True)
+SERVER_ID = 813324385179271168
+# Testing Server
+# SERVER_ID = 836589565237264415
+
+MESSAGE_ID = 950265978888007770
```
TomCranitch, line 12: Should this come from joinmessage()

uqcsbot/join.py
```diff
@@ -0,0 +1,103 @@
...
+        channel_emojis = {}
+        for channel in channel_query:
+            if channel.name in EMOJIS:
+                channel_emojis[channel.name] = EMOJIS[channel.name]
+        return channel_emojis
+
+    @commands.Cog.listener()
+    async def on_raw_reaction_add(self, payload):
+        """ Toggle adding/removing member from the corresponding channel. """
+        channels = self.get_channel_map()
```
TomCranitch, line 52: this should probably only be done inside the if statement because this function is reasonably expensive and we dont want it to run every time a reaction is added

uqcsbot/join.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @commands.command(hidden=True)
+    @commands.has_permissions(manage_channels=True)
+    async def joinmessage(self, ctx: commands.Context):
+        """ Create message to react to. """
+        channels = self.get_channel_map()
+        channel_list = list(channels.items())
+        message = "**Channel Menu:**\nReact to join these channels.\n\n"
+
+        for name, emoji in channel_list:
```
TomCranitch, line 93: This doesn't seem to work on the first run (as the channel_map will be empty)

TomCranitch (commented):
Thanks for the PR it looks pretty solid. That being said it seems it doesn't work at the moment. I'm not sure if I'm missing a step because it looks like you had it working in the testing server

image

I've added a comment on line 93 with what I think is causing the issue. I do wonder if we need to even store the channels in the db or if instead when a reaction is added the bot could lookup the channel ID based on it's name. The issue with this is that we would require unique channel names, so I'm not sure if anyone else has any strong opinions either way

uqcsbot/join.py
```diff
@@ -0,0 +1,103 @@
...
+JOINED_PERMISSIONS = discord.Permissions(read_messages=True)
+SERVER_ID = 813324385179271168
+# Testing Server
+# SERVER_ID = 836589565237264415
+
+EMOJIS = {"academic-advice": "📖", "adulting": "🧑", "banter": "💬", "bot-testing" : "🤖",
+                "contests" : "⚔️", "covid" : "⚛️", "creative" : "🎨", "events" : "🗓️", "food" : "🍔",
+                "games" : "🎮", "general" : "⚪", "hackathons" : "🍕", "hardware" : "💻", "jobs-bulletin" : "📌",
+                "jobs-discussion" : "🔈", "lgbtqia" : "🏳️‍🌈", "media" : "📺", "memes" : "🎭", "politics" : "📮",
+                "projects" : "🔨", "uqic-sport" : "🏅", "yelling" : "🗣️"}
```
JamesDearlove, line 16: Building upon what we do already have, would it be beneficial to store these alongside the channels within the database already? Also means we don't have to constantly update the code if we add/remove channels from the Discord.

uqcsbot/join.py
```diff
@@ -0,0 +1,103 @@
+import discord
```
JamesDearlove, line 1: Is creating a separate cog ideal for this? We do already have the channels cog, which I think these should be a part of. Also should help reduce a bit of code duplication

uqcsbot/join.py
```diff
@@ -0,0 +1,103 @@
...
+                "contests" : "⚔️", "covid" : "⚛️", "creative" : "🎨", "events" : "🗓️", "food" : "🍔",
+                "games" : "🎮", "general" : "⚪", "hackathons" : "🍕", "hardware" : "💻", "jobs-bulletin" : "📌",
+                "jobs-discussion" : "🔈", "lgbtqia" : "🏳️‍🌈", "media" : "📺", "memes" : "🎭", "politics" : "📮",
+                "projects" : "🔨", "uqic-sport" : "🏅", "yelling" : "🗣️"}
+
+class Join(commands.Cog):
+
+    def __init__(self, bot: UQCSBot):
+        self.bot = bot
+        self.message_id = None
```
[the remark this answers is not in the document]
JamesDearlove: Important as we have no guarantees for long long the bot will be running, and is rebooted on new deployments

uqcsbot/join.py
JamesDearlove: This should work if you have channels in the "old" channel list

JamesDearlove (changes requested):
Solid good work here, I like this approach as it addresses most of my concerns around it and makes it clearer to new members. Most of my comments are similar to Tom's.

Main thing I'd do here is combine your code into the `channels` cog as its all related. Also taking advantage of the DB here for storing emojis would be good.

JamesDearlove (changes requested):
Solid good work here, I like this approach as it addresses most of my concerns around it and makes it clearer to new members. Most of my comments are similar to Tom's.

Main thing I'd do here is combine your code into the `channels` cog as its all related. Also taking advantage of the DB here for storing emojis would be good.

uqcsbot/join.py
```diff
@@ -0,0 +1,103 @@
...
+        db_session.close()
+
+        channel_emojis = {}
+        for channel in channel_query:
+            if channel.name in EMOJIS:
+                channel_emojis[channel.name] = EMOJIS[channel.name]
+        return channel_emojis
+
+    @commands.Cog.listener()
+    async def on_raw_reaction_add(self, payload):
```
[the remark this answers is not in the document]
campbell-mcfadden: Thanks for the feedback. The old way to join channels will ideally get phased out eventually, however I will look again to see any potential for shared functionality between the two

uqcsbot/join.py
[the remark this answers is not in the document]
campbell-mcfadden: Fixed in [c72ec9a](https://github.com/UQComputingSociety/uqcsbot-discord/pull/43/commits/c72ec9a04952f82af1d934b126421c41f51c246a)

uqcsbot/join.py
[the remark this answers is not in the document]
campbell-mcfadden: I had not thought of this actually, this is a good catch on your part. I will look at keeping this value stored in the db instead

uqcsbot/join.py
campbell-mcfadden: Thanks for the review. That is definitely a much better long-term solution. I will look at updated the db to store these values instead of inside the code itself

uqcsbot/join.py
campbell-mcfadden: My thoughts here were that given ``!joinchannel`` and ``!leavechannel`` are going to be phased out eventually, it would be easier just to separate them from this new feature. However it is definitely ideal to have all the joining/leaving centralised to one cog for the time being so I will merge them in the next commit

uqcsbot/join.py
```diff
@@ -0,0 +1,103 @@
...
+        self.message_id = None
+
+    def _channel_query(self, channel: str):
+        db_session = self.bot.create_db_session()
+        channel_query = db_session.query(Channel).filter(Channel.name == channel,
+                                                         Channel.joinable == True).one_or_none()
+        db_session.close()
+        return channel_query
+
+    def get_key(self, map, value):
```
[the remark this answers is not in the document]
campbell-mcfadden: Given the key gets indexed more often, I thought it would be okay to leave the map in the current order. Although this is definitely a method that can be replaced with a one-liner

uqcsbot/join.py
```diff
@@ -0,0 +1,103 @@
...
+
+            # Remove reaction if not a bot
+            if not member.bot:
+                await msg.remove_reaction(payload.emoji, member)
+
+            channel = self.get_key(channels, payload.emoji.name)
+            channel_query = self._channel_query(channel)
+
+            if channel_query == None:
+                await member.send(f"Unable to find that channel.")
```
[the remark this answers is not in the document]
campbell-mcfadden: Thanks, have added that in

campbell-mcfadden:
I'm going to have another go at this, will look to store emojis in the DB as well as merge the functionality between channels.py and join.py

JamesDearlove:
Looking forward to reviewing your updated code :)

In regards to the old system, I personally don't see why the old commands can't stick around. If they use the same base functionality, its just another path for people to join channels.

Issue #2, opened 2021-06-05
Testing Suite


JamesDearlove:
Off the back of #1, it would be good to have a testing suite in place. Seperate issue created for discussion.

49Indium:
How would we like tests to work? There is a test folder currently; do we want to have tests for all components?

JamesDearlove:
@49Indium Ideally we should test things that are within our control. For example how the events are parsed from the ical, rather than testing if a message is sent.

49Indium:
I've added tests (that pass) for Haiku. Will add some for the other cogs soon.

49Indium:
Looking at this further, it seems that we might need to relocate some of the logic of some commands to their own functions (not within the object) to be able to test them effectively without simulating discord (see [here](https://github.com/Rapptz/discord.py/issues/197)).

andrewj-brown:
linking this library for if/when we ever get the time. This looks to be the "best" (most up-to-date and still live) discord mocking library: https://github.com/CraftSpider/dpytest

PR #44, opened 2022-05-15
react-joins #2

campbell-mcfadden:
This adds an epic™ new way of joining/leaving channels by reacting to a uqcsbot message. The message updates itself automatically when new channels are made joinable, and users are DM'd when they join or leave. It should help reduce some of the spam in the `#bot-testing` channel.

This time around I used a lot of the pre-existing infra to store the reactions and the message ID. Let me know what y'all think, I'm open to any suggestions.

### DIFF
--- a/uqcsbot/channels.py
+++ b/uqcsbot/channels.py
@@ -7,10 +7,14 @@ from sqlalchemy.exc import NoResultFound
 from uqcsbot.bot import UQCSBot
 from uqcsbot.models import Channel
 
+from emoji import UNICODE_EMOJI_ENGLISH
+
 JOINED_PERMISSIONS = discord.Permissions(read_messages=True)
 SERVER_ID = 813324385179271168
 # Testing Server
-# SERVER_ID = 836589565237264415
+#SERVER_ID = 836589565237264415
+
+MESSAGE_ID = 974553937984229386
 
 class Channels(commands.Cog):
 
@@ -24,6 +28,10 @@ class Channels(commands.Cog):
         db_session.close()
 
         return channel_query
+    
+    def _valid_emoji(self, emoji):
+        custom = [f"<:{e.name}:{e.id}>" for e in self.bot.emojis]
+        return emoji in UNICODE_EMOJI_ENGLISH or emoji in custom
 
     @commands.command()
     async def joinchannel(self, ctx: commands.Context, *channels: str):
@@ -61,7 +69,7 @@ class Channels(commands.Cog):
     @commands.command()
     async def leavechannel(self, ctx: commands.Context, channel=""):
         """ Leaves the channel that you specify. """
-
+        
         # If a channel is not specified, attempt to leave the current channel.
         if (channel == ""):
             channel = ctx.channel.name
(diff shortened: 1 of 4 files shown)


uqcsbot/channels.py
```diff
@@ -109,36 +126,67 @@ async def listchannels(self, ctx: commands.Context):
...
-
         db_session.commit()
         db_session.close()
+
         await ctx.send(f"{channel.mention} was removed as a joinable channel.")
+        react_message = await ctx.fetch_message(self.message_id)
+        await self.update_message(react_message)
+
+    @commands.command(hidden=True)
+    @commands.has_permissions(manage_channels=True)
```
JamesDearlove, line 175: Making this hidden hides it for all users, including committee. I'd leave it not hidden as it will automatically be hidden from help for anyone who can't manage channels.

uqcsbot/channels.py
```diff
@@ -109,36 +126,67 @@ async def listchannels(self, ctx: commands.Context):
 
     @commands.command()
     @commands.has_permissions(manage_channels=True)
-    async def addjoinchannel(self, ctx: commands.Context, channel: discord.TextChannel):
+    async def addjoinchannel(self, ctx: commands.Context, channel: discord.TextChannel, emoji):
```
JamesDearlove, line 129: Minor issue: `emoji` should ideally have a type, I think it being `str` should be fine

uqcsbot/channels.py
```diff
@@ -149,5 +197,62 @@ async def channel_manage_error(self, ctx: commands.context, error):
...
+
+            # Leave the channel if the user is currently a member.
+            if channel.permissions_for(member).is_superset(JOINED_PERMISSIONS):
+                await channel.set_permissions(member, read_messages=False, reason="UQCSbot removed.")
+                await member.send(f"You've left {channel.mention}")
+                return
+
+            # Otherwise, join the channel.
+            await channel.set_permissions(member, read_messages=True, reason="UQCSbot added.")
+            await member.send(f"You've joined {channel.mention}")
```
JamesDearlove, line 240: Unsure if we necessarily need do DM the user every time they join and leave a channel. This was a point brought up a while ago as it may get annoying?

JamesDearlove (changes requested):
Nice work, and apologies that I'm only getting around to this now. 

One major concern I have regarding the model in a comment below. Plus some other notes, but looks pretty good imo.

uqcsbot/channels.py
campbell-mcfadden: Oh true, I hadn't considered it would be invisible for committee as well. Thanks.

uqcsbot/channels.py
campbell-mcfadden: I see your side of things there, however committee seems pretty set on the idea of having dms as confirmation. Since joining/leaving channels is a rare enough operation I think it is fine, but we can easily repeal this if reception is poor ;)

uqcsbot/models.py
```diff
@@ -9,14 +9,20 @@ class Channel(Base):
     id = Column("id", BigInteger, primary_key=True, nullable=False)
     name = Column("name", String, nullable=False)
     joinable = Column("joinable", Boolean)
+    emoji = Column("emoji", String, nullable=False)
```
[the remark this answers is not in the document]
campbell-mcfadden: Hmm, since I manually modified this table with sql would any further migrations be necessary do you think? Things seem to be okay in my test env. Happy to take at look at using Alembic for the schema migration if more work is needed. Thanks.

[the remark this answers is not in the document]
JamesDearlove: I think for now that should be fine, Alembic and schema migrations are not a small task and I think this is probably a one off event anyway.

uqcsbot/channels.py
JamesDearlove: If that is what committee decree, who am I to argue

JamesDearlove (approved):
lgtm, this will be nice to finally have

PR #45, opened 2022-05-21
No more yelling at links (plus a little docstring fix)

leadnaut:
Docstring was incorrect for function

Regex added to prevent links from being yelled at in the same method as the current method for ignoring emojis.

I'm not a regex pro so I stole the expression from: https://stackoverflow.com/questions/21932615/regular-expression-for-remove-link


Thomas-Malcolm:
Updated the regex to identify URL encoding as well

PR #46, opened 2022-05-21
#YELLING Regex Updates

Thomas-Malcolm:
Certainly not exhaustive, please add if there's more you know of - this at least should cover most common use cases


JamesDearlove (approved):
lgtm

PR #52, opened 2022-10-18
Proper nouns require proper capitalisation

ThatSealgair:
Given that this function has been named after and developed to deter a specific individual, it would be best to highlight the function's importance by utilising proper capitalization of the offender's name to prevent later removal.


bradleysigma:
Consistent code styling takes precedent over normal English grammar rules; if Caesar doesn't get capitalised, then you surely don't.

PR #53, opened 2022-10-28
Update text.py

ThatSealgair:
No dox pls


ThatSealgair:
Pls sir, no dox I will be good man.

PR #63, opened 2022-12-12
Update gaming.py

bradleysigma:
Fixes double faced cards for `/scry`.


JamesDearlove (approved):
While not elegant user side, this does indeed fix the problem

PR #64, opened 2022-12-13
More app commands

andrewj-brown:
Added "app commands" for zalgo, mock, and scare. Also added /smoko.

This is in lieu of the long-threatened "reply" idea, where the bot would mock/scare/zalgo a replied message - now, we can just use the app commands menu instead.

Also, this includes the current installment of @JamesDearlove's dev-improvements branch.


JamesDearlove (approved):
lgtm, and thanks for pulling those changes in (and testing them)

Issue #57, opened 2022-11-14
Improvements to #yelling


JamesDearlove:
Some people are finding ways around our sacred #yelling channel rules, and should be fixed up. The two potential improvements are:
- [x] Detecting messages being edited, and alerting the channel
- [ ] Potentially blocking txt attachments
- [ ] Zalgo command
- [x] Consequences for repeated violations of the ALL CAPS policy

ew-b:
Love it, let’s slap on a cool down mode as a consequence 🥰🥰

bradleysigma:
I've been thinking about doing this, once the bot is updated to discord.py v2. Does the bot have mod permissions?

JamesDearlove:
It should, and if it doesn't we can fix that

peclarke:
one god, one bot. Give it ownership of the discord

andrewj-brown:
the mod-like perms that uqcsbot currently has are:
- manage roles (only self-assignable + tutor role though)
- mention everyone, here, & all roles
- manage messages
- manage threads

It may need "manage channels" to enforce yelling via slowmode, but nothing else. I support the idea of keeping uqcsbot's privileges as minimal as possible and only adding when there's a clear use-case (e.g. mention everyone is something I don't think the bot should have, because there's no use-case, but I've been overruled on that).

andrewj-brown:
new, semi-related improvement to add to this: /echo should convert messages to uppercase if they occur in #yelling

49Indium:
Bradley is working on this with #152 and #153

PR #65, opened 2022-12-16
Update basic.py

bradleysigma:
Prevent using `/echo` to avoid yelling in `#yelling`.


andrewj-brown:
I had thought of this, but intended to make uqcsbot simply .upper() the message before replying, which is possibly better than just replacing the entire message with disapproval.

This would revoke the need for my comment on #57 if implemented.

JamesDearlove (changes requested):
So we aren't beholden to the ID system going forward, can you change this to use the `discord.utils.get` system?

Example here: https://github.dev/UQComputingSociety/uqcsbot-discord/blob/main/uqcsbot/voteythumbs.py#L23

JamesDearlove (approved):

PR #40, opened 2022-03-04
Minecraft Server Additions

JamesDearlove:
With the introduction of the Minecraft server, these add in two useful commands.

`mcwhitelist` will let anyone within the UQCS Discord add someone to the whitelist.
`mcadmin` allows people with the Manage Server role (should be only committee + me currently) to run commands directly on the server.

These changes add in some additional environment variables so these can be configured without code change, and also not "leaking" the server URL to the rest of the world. 
`MC_RCON_ADDRESS` Address or IP address of the server with enabled RCON.
`MC_RCON_PORT` The port used for connecting to RCON.
`MC_RCON_PASSWORD` The password configured for RCON connection.

These changes require that the Minecraft server to have RCON enabled in `server.properties`, requires PR #39 merged for the bot to be running in Azure, and the Minecraft server and UQCSbot to be communicating privately.

Suggestions and feedback welcome, I'm still not entirely sure if anyone in the UQCS Discord should just be able to add people to the whitelist. I leave that for committee to decide.

PR #66, opened 2022-12-22
Commands for Minecraft Server

JamesDearlove:
An upgraded and improved version of #40 

`/mcwhitelist` will let a UQCS member add one user to the whitelist. Admins override the one user limit
`/mcadmin run` allows people with correct permissions to manage the server directly. More can be added to this in the future so people don't have to remember specific commands.

Additionally, a new bot wide function to log directly to admin-alerts. Whitelisting will add when successful, and all runs of `mcadmin` are logged.

Required new environment variables for this to work:
`MC_RCON_ADDRESS` Address or IP address of the server with enabled RCON.
`MC_RCON_PORT` The port used for connecting to RCON.
`MC_RCON_PASSWORD` The password configured for RCON connection.


LimaoC (approved):
Looks good to me!

Once this is merged, we should probably clear the current whitelist as it's already got a decent number of players on it.

If a member wants to un-whitelist/change the name on the whitelist, I imagine they'd just contact committee for that? An un-whitelist command would probably be pretty abusable.

Also at some point we'll likely lock this command behind some sort of verified UQCS member role.

JamesDearlove:
We can add some admin commands to remove people from the whitelist in the future, thought it would be more beneficial to get this up and running now.

uqcsbot/minecraft.py
```diff
@@ -0,0 +1,115 @@
...
+
+        return response
+
+    @app_commands.command()
+    @app_commands.describe(username="Minecraft username to whitelist.")
+    async def mcwhitelist(self, interaction: discord.Interaction, username: str):
+        """ Adds a username to the whitelist for the UQCS server. """
+        db_session = self.bot.create_db_session()
+        query = db_session.query(MCWhitelist).filter(MCWhitelist.discord_id == interaction.user.id)
+        is_user_admin = interaction.user.guild_permissions.manage_guild
```
LimaoC, line 50: There is a slight caveat to this, `interaction.user` can either be of type `User` or `Member`, and only `Member` has the `guild_permissions` field. I *think* this is to do with whether you send the bot a command in the server or not, as it won't be able to check your permissions without knowing which server you're talking about.

JamesDearlove: Ah good catch, I'll put a check in for this at some point

andrewj-brown:
mc server hella dead, is it even worth adding this to the bot when no-one uses it?

PR #69, opened 2023-01-29
Add Reminders

LimaoC:
Adds a "remind me" functionality to the bot. The intended purpose is for committee to set "secondary" announcements in relevant channels (CAHP due dates in the tutor channel, CPG reminders in the cpg channel, etc.), though regular members may also set reminders. Members have a limit of 10 reminders (configurable in `remindme.py`), and mentions are disabled in reminder messages to prevent abuse. Admins override both of these restrictions.

A brief overview of the available commands, wrapped under the cog `RemindMe`:
- `/remindme add` adds a one-time reminder.
- `/remindme addrecurring` adds a recurring reminder.
- `/remindme remove` removes an existing reminder belonging to the user.
- `/remindme list` lists all active reminders belonging to the user.

### DIFF
--- a/uqcsbot/remindme.py
+++ b/uqcsbot/remindme.py
@@ -4,7 +4,6 @@ from discord import app_commands
 from discord.ext import commands
 from functools import partial
 import logging
-import random
 from typing import List, NamedTuple, Optional, Union
 
 from uqcsbot.bot import UQCSBot
@@ -12,7 +11,6 @@ from uqcsbot.models import Reminders
 
 
 USER_REMINDER_LIMIT = 10
-REMINDER_IDS_RANGE = (1, 100)  # range from which reminder ids will be randomly chosen
 
 REMINDME_MESSAGE_TITLE = "RemindMe: Reminder"
 REMINDME_ERROR_TITLE = "RemindMe: Error"
@@ -182,12 +180,13 @@ class RemindMe(commands.Cog):
         """ Returns all the active reminders belonging to the user with id `user_id` in creation order """
         return list(filter(lambda reminder: reminder.user_id == user_id, self._get_all_reminders()))
     
-    def _get_unused_reminder_id(self, user_id: int) -> int:
-        """ Returns a reminder id that is not currently in use by the user with id `user_id` """
-        reminders = self._get_user_reminders(user_id)
+    def _get_unused_reminder_id(self) -> int:
+        """ Returns a reminder id that is not currently in use """
+        reminders = self._get_all_reminders()
         reminder_ids = [reminder.id for reminder in reminders]
-        while (id := random.randint(*REMINDER_IDS_RANGE)) in reminder_ids:
-            continue
+        i = 1
+        while (id := i) in reminder_ids:
+            i += 1
         return id
 
     def _reached_reminder_limit(self, user: Union[discord.User, discord.Member]) -> bool:
@@ -313,7 +312,7 @@ class RemindMe(commands.Cog):
 
         # add reminder to db and schedule
         reminder = Reminder(
-            self._get_unused_reminder_id(interaction.user.id),
+            self._get_unused_reminder_id(),
             interaction.user.id,
             interaction.channel_id,
             dt.datetime.now(),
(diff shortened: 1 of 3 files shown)


JamesDearlove (commented):
No glaring issues, I haven't tested it, but it seems to be all good. lgtm

LimaoC:
Just remembered - reverted the DB engine setup & cog loading order just to be safe, pre-existing reminders will be scheduled once the bot is ready

JamesDearlove (approved):

uqcsbot/remindme.py
```diff
@@ -0,0 +1,408 @@
...
+            reminder.start_date,
+            reminder.end_date,
+            reminder.week_frequency
+        ) for reminder in sorted(reminders_query, key=lambda reminder: reminder.time_created)]
+
+    def _get_user_reminders(self, user_id: int) -> List[Reminder]:
+        """ Returns all the active reminders belonging to the user with id `user_id` in creation order """
+        return list(filter(lambda reminder: reminder.user_id == user_id, self._get_all_reminders()))
+    
+    def _get_unused_reminder_id(self, user_id: int) -> int:
```
andrewj-brown, line 185: Random seems like significantly more effort than just incremental when you have to factor in the chance of duplicates, and comes with a risk of extremely high response times - worst case is ~1/10'th chance of duplicate, which is low, but not zero

uqcsbot/remindme.py
```diff
@@ -0,0 +1,408 @@
...
+    channel_id: int | None  # None => send reminder in DMs
+    time_created: dt.datetime
+    message: str
+    time: dt.time
+    start_date: dt.date
+    end_date: dt.date | None  # None => non-ending
+    week_frequency: int | None  # 0 => daily, None => one-time
+
+    def __str__(self):
+        id = self.id
```
andrewj-brown, line 52: Use of reminder id here - outputting reminders publicly will display a lot of duplicate IDs, which isn't a problem internally but may confuse people

uqcsbot/remindme.py
```diff
@@ -0,0 +1,408 @@
...
+def _error_embed(description: str, footer: str = "") -> discord.Embed:
+    """ Returns a formatted Error embed with the given description and an optional footer """
+    embed = discord.Embed(title=REMINDME_ERROR_TITLE, description=description, color=discord.Color.red())
+    if footer != "":
+        embed.set_footer(text=footer)
+    return embed
+
+def _add_reminder_embed(reminder: Reminder, footer: str = "") -> discord.Embed:
+    """ Returns a formatted Reminder Added embed with the given Reminder and an optional footer """
+    embed = discord.Embed(title=REMINDME_ADDED_TITLE, description=str(reminder))
```
andrewj-brown, line 87: 2310 moment: defining all these instead of just using them in the one place they crop up :joy: (not a problem, just memeing)

uqcsbot/remindme.py
```diff
@@ -0,0 +1,408 @@
...
+        reminders = self._get_user_reminders(user_id)
+        reminder_ids = [reminder.id for reminder in reminders]
+        while (id := random.randint(*REMINDER_IDS_RANGE)) in reminder_ids:
+            continue
+        return id
+
+    def _reached_reminder_limit(self, user: Union[discord.User, discord.Member]) -> bool:
+        """ Returns whether the given user has reached the maximum reminder limit. """
+        # a slight caveat is that if a user sends a command to the bot in DMs, it won't be able to
+        # check for admin perms, so the reminder limit will be enforced
```
andrewj-brown, line 196: Theoretically, because the bot is only in one guild, it could query the guild for the user id and check perms, but that's a lot of effort, so... eh

andrewj-brown (commented):
most of my comments are just "huh", not merge-blocking problems

uqcsbot/remindme.py
LimaoC: Yep this is a fair point - I was originally bothered by having non-ending incremental ids, but since there's a reminder limit the id can always just be chosen between `(0, USER_REMINDER_LIMIT)`

uqcsbot/remindme.py
LimaoC: That's true, it would probably be clearer for people if I made ids unique across all reminders rather than just for a specific user. I was also thinking of implementing admin commands at some point (so admins can remove others' reminders), so having unique ids across the board would make that easier as well.

uqcsbot/remindme.py
LimaoC: Ehhh yeah too much effort :joy:

PR #73, opened 2023-03-08
Holidays

LimaoC:
Ports over the holiday cog from `uqcsbot-slack`.

Every day, the UQCS bot will pick a random holiday occuring on that day and send `"Today is {holiday}!"` at 9am to #general.

### DIFF
--- a/uqcsbot/holidays.py
+++ b/uqcsbot/holidays.py
@@ -1,15 +1,21 @@
-from uqcsbot import bot
-from uqcsbot.utils.command_utils import HYPE_REACTS
 from bs4 import BeautifulSoup
-from datetime import datetime
-from random import choice
-from requests.exceptions import RequestException
-import requests
 import csv
+from datetime import datetime
+import discord
+from discord.ext import commands
+import logging
+from random import choice
+import requests
+from requests.exceptions import RequestException
 from typing import List
+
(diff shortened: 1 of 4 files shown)


uqcsbot/holidays.py
```diff
@@ -94,14 +76,45 @@ def get_holidays_from_csv() -> List[Holiday]:
...
         return response.content
     except RequestException as e:
-        bot.logger.error(e.response.content)
+        logging.warning(e.response.content)
     return None
+
+class Holidays(commands.Cog):
+    def __init__(self, bot: UQCSBot):
+        self.bot = bot
+        self.bot.schedule_task(self.holiday, trigger='cron', hour=16, minute=12, timezone='Australia/Brisbane')
```
JamesDearlove, line 93: Might need to actually be 9am

```suggestion
        self.bot.schedule_task(self.holiday, trigger='cron', hour=9, minute=0, timezone='Australia/Brisbane')
```

JamesDearlove (changes requested):
Just a minor (ok maybe not minor) issue

uqcsbot/holidays.py
LimaoC: Ah that is a big oopsies, thanks

PR #76, opened 2023-03-27
Make whatweekisit timezone-aware

TheOnlyMrCat:
I thought this would be 2 lines, but it's actually 3. Please forgive me.

This makes the `/whatweekisit` command say today's date, rather than yesterday's, when run between the hours of 
00:00 and 10:00 (Brisbane time).

### DIFF
--- a/uqcsbot/whatweekisit.py
+++ b/uqcsbot/whatweekisit.py
@@ -2,6 +2,7 @@ import discord
 from discord import app_commands
 import requests
 from typing import NamedTuple, List, Tuple, Optional
+from zoneinfo import ZoneInfo
 from datetime import datetime, timedelta
 from math import ceil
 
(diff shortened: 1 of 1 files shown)


uqcsbot/whatweekisit.py
```diff
@@ -2,6 +2,7 @@
 from discord import app_commands
 import requests
 from typing import NamedTuple, List, Tuple, Optional
+from zoneinfo import ZoneInfo
```
andrewj-brown, line 5: in theory you should be able to get away with using the existing datetime imports: replace your ZoneInfo call with `timezone(timedelta(hours=10), "AEST")`

andrewj-brown (approved):

uqcsbot/whatweekisit.py
JamesDearlove: Gonna disagree as I do prefer the `Australia/Brisbane` definition as it's a tiny bit clearer (alas an extra import, but we already do this for pytz.

JamesDearlove (approved):
I'm happy with this approach, lets us also get away from having the pytz library as well. Also learnt today that `zoneinfo` is actually a thing in Python 3.9

PR #79, opened 2023-04-06
Update reminders db to use BigIntegers for user and channel ids

LimaoC:


JamesDearlove (approved):
This will probably require a manual drop of the table on the db once merged

PR #80, opened 2023-04-06
Added Cowsay

lcox74:
Had a weird thought, "what if `cowsay` was in discord?" Then I accidentally built that instead of doing any work.

The command is `/cowsay message`

```
 _____________ 
< hello world >
 ------------- 
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\ 
                ||----w |
                ||     ||
```

There are also optional arguments:

- `tongue` which will allow the cow to stick out its tongue
- `tux` which changes the cow to the Linux Tux
- `mood` which will change the eyes from a select list:

```py
select_list = dict(
    Normal = 'oo',
    Borg = '==',
    Dead = 'xx',
    Greed = '$$',
    Paranoid = '@@',
    Stoned = '**',
    Tired = '--',
    Wired = 'OO',
    Youthful = '..'
)
```

### DIFF
--- a/uqcsbot/cowsay.py
+++ b/uqcsbot/cowsay.py
@@ -1,6 +1,5 @@
-from random import choice
 
-from typing import List, NamedTuple, Optional, Union, Literal
+from typing import Optional, Literal
 
 import discord
 from discord import app_commands
@@ -8,6 +7,8 @@ from discord.ext import commands
 
 from uqcsbot.bot import UQCSBot
 
+# Type Alias for Cow Eyes 
+CowsayMoodType = Literal['Normal', 'Borg', 'Dead', 'Greed', 'Paranoid', 'Stoned', 'Tired', 'Wired', 'Youthful']
 
 class Cowsay(commands.Cog):
 
@@ -35,7 +36,7 @@ class Cowsay(commands.Cog):
     )
     async def cowsay_command(self, interaction: discord.Interaction, 
             message: str, 
-            mood: Optional[Literal['Normal', 'Borg', 'Dead', 'Greed', 'Paranoid', 'Stoned', 'Tired', 'Wired', 'Youthful']] = 'Normal', 
+            mood: Optional[CowsayMoodType] = 'Normal', 
             tongue: Optional[bool] = False,
             tux: Optional[bool] = False) -> None:
         """
@@ -88,7 +89,7 @@ class Cowsay(commands.Cog):
         return bubble
 
     def draw_cow(self, 
-            mood: Optional[Literal['Normal', 'Borg', 'Dead', 'Greed', 'Paranoid', 'Stoned', 'Tired', 'Wired', 'Youthful']] = 'Normal', 
+            mood: Optional[CowsayMoodType] = 'Normal', 
             tongue: Optional[bool] = False) -> str:
 
         """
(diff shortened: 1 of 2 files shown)


uqcsbot/cowsay.py
```diff
@@ -0,0 +1,135 @@
...
+        
+        # Construct the speech bubble.
+        bubble = f" _{bubble_length * '_'}_ \n"
+        bubble += bubble_body
+        bubble += f" -{bubble_length * '-'}- \n"
+
+        return bubble
+
+    def draw_cow(self, 
+            mood: Optional[Literal['Normal', 'Borg', 'Dead', 'Greed', 'Paranoid', 'Stoned', 'Tired', 'Wired', 'Youthful']] = 'Normal',
```
JamesDearlove, line 91: I think as this type is used a few more times, you should have a type alias for it.

uqcsbot/cowsay.py
```diff
@@ -0,0 +1,135 @@
+from random import choice
```
JamesDearlove, line 1: Nit: Unused import

uqcsbot/cowsay.py
```diff
@@ -0,0 +1,135 @@
...
+
+from typing import List, NamedTuple, Optional, Union, Literal
```
JamesDearlove, line 3: Nit: NamedTuple and Union are also unused

JamesDearlove (changes requested):
Thanks for contributing! Hope you didn't procrastinate your work for too long. Nice one, didn't expect we'd need this, but yeah we absolutely do.

Just a couple of minor changes

lcox74:
I amended the code to reflect the requested changes. 

Im currently out camping, writing code on your phone sucks. Also can't test to see if aliasing the Literal type works, python be pedantic.

uqcsbot/cowsay.py
```diff
@@ -0,0 +1,136 @@
+
+from typing import Optional, Literal
```
JamesDearlove, line 2: Missing List from typing
```suggestion
from typing import Optional, Literal, List
```

uqcsbot/cowsay.py
```diff
@@ -0,0 +1,141 @@
...
+            message: str, 
+            mood: Optional[CowsayMoodType] = 'Normal', 
+            tongue: Optional[bool] = False,
+            tux: Optional[bool] = False) -> None:
+        """
+        Returns a cow or tux saying the given message.
+        """
+
+        # Check message length, if invalid send moo!
+        if len(message) == 0 or len(message) > 2000:
```
JamesDearlove, line 47: Cap this at 1000 characters, as its the final message it sends back that needs to be less than 2000 characters.
```suggestion
        if len(message) == 0 or len(message) > 1000:
```

uqcsbot/cowsay.py
```diff
@@ -0,0 +1,141 @@
...
+            Paranoid = '@@',
+            Stoned = '**',
+            Tired = '--',
+            Wired = 'OO',
+            Youthful = '..'
+        )
+
+    @app_commands.command(name="cowsay")
+    @app_commands.describe(
+        message="The message that the cow will say",
```
JamesDearlove, line 32: Just making it clear that the messsage can only contain 1000 characters.

```suggestion
        message="The message that the cow will say (Max 1000 chars)",
```

JamesDearlove (approved):
Approved pending these final suggestions

PR #81, opened 2023-04-07
add syllable-check command

andrewj-brown:
add quick syllable-check command to ~~stop 17-word haiku attempts~~ make 17-word haiku attempts not the only way to test stuff

### DIFF
--- a/uqcsbot/haiku.py
+++ b/uqcsbot/haiku.py
@@ -47,7 +47,12 @@ class Haiku(commands.Cog):
     @app_commands.command()
     @app_commands.describe(text="Syllable check for a given word")
     async def syllables(self, interaction: discord.Interaction, word: str):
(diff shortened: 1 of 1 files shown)


uqcsbot/haiku.py
```diff
@@ -43,6 +44,11 @@ async def on_message(self, message):
         else:
             await message.reply(f"Nice haiku:\n{haiku}")
 
+    @app_commands.command()
+    @app_commands.describe(text="Syllable check for a given word")
```
JamesDearlove, line 48: This should refer to the word arg, instead of text

```suggestion
    @app_commands.describe(word="Syllable check for a given word")
```

JamesDearlove (changes requested):
Two things:
1. Can you add a comment to the command, so it gets populated in the command list.
2. Minor issue with the .describe

JamesDearlove (approved):
lgtm

PR #84, opened 2023-04-10
Added /membercount and an auto-updated VC to show member numbers

49Indium:
The /membercount command displays the current member, count (updated to when the command was called) along with the number of users which joined in the last 7 days.
The VC (which should be visible but not accessible for most users), has its named changed when someone leaves or joins the server, but is capped at refreshing at most every 5 minutes (as the current rate limit is 2 channel name changes per 10 minutes).


uqcsbot/member_counter.py
```diff
@@ -0,0 +1,79 @@
...
+import asyncio
+
+import discord
+from discord import app_commands
+from discord.ext import commands
+
+
+class MemberCounter(commands.Cog):
+    # Checks for an Azure specific environment variable, if it exists we're running as prod.
+    MEMBER_COUNT_CHANNEL_ID = 1094584746207617064 if os.environ.get("WEBSITE_SITE_NAME") != None \
```
andrewj-brown, line 13: We want to not use IDs as much as possible, because they're opaque to any readers and require funky hacks like this to remain available in other servers (consider: the bot is locally hostable by anyone in any server, not just uqcs & testing). Is it possible to just get all channels and filter by "name starts with Member Count" ?

This is the one case where IDs are justifiable, due to name changes, so don't worry about it too much.

andrewj-brown (commented):
I'll defer to other reviewers on whether the IDs are a blocking issue in this PR, but I'd like to see them changed if possible. The name changes make this harder than other use-cases though so the mileage may vary.

uqcsbot/member_counter.py
49Indium: Modified to use the beginning of the channel name to determine the right channel.

On a related note, should `SERVER_ID` within `bot.py` be changed as well? Do we guarantee that uqcsbot is only in one server (as if this is not true, grabbing channels by name is dubious).

andrewj-brown: uhhhhhhhhhhhhhhhhh good question

andrewj-brown: alright I may consistent-ise that shortly but for now if you just use the standard approach here (get_all_channels() and filter) then that should be fine

andrewj-brown (approved):
lgtm

49Indium:
Given the recent changes, I'll let someone review again. Sorry for the hassle.

andrewj-brown (approved):
lgtm still. dunno about ending cogs; just blocking further updates should be ok (the only case where it "doesn't work" is an incorrectly configured server, which shouldn't happen in prod)

andrewj-brown:
keep in mind you should setup the channel & perms in the discord server before merging this

PR #86, opened 2023-04-10
Modified uptime to also work in test server

49Indium:
Uptime now finds #bot-testing by using a different channel ID depending on a Azure specific environment variable. This approach is the same as used in `bot.py`, so should have no issues. This should (hopefully) not affect production and provide reboot messages to the testing server.

### DIFF
--- a/uqcsbot/uptime.py
+++ b/uqcsbot/uptime.py
@@ -11,21 +11,19 @@ from uqcsbot.bot import UQCSBot
 
 
 class UpTime(commands.Cog):
-    # Checks for an Azure specific environment variable, if it exists we're running as prod.
-    CHANNEL_ID = 836243768411160606 if os.environ.get("WEBSITE_SITE_NAME") != None \
-        else 1041708952066461716
+    CHANNEL_NAME = "bot-testing"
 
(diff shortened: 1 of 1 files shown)


uqcsbot/uptime.py
```diff
@@ -10,7 +11,9 @@
 
 
 class UpTime(commands.Cog):
-    CHANNEL_ID = 836243768411160606
+    # Checks for an Azure specific environment variable, if it exists we're running as prod.
+    CHANNEL_ID = 836243768411160606 if os.environ.get("WEBSITE_SITE_NAME") != None \
```
andrewj-brown, line 15: Similar to my comments on /membercount, we should avoid using IDs wherever possible. See other cogs (e.g. starboard) for name-based handling of channels - it needs an on_ready handler, but in return, it works for anyone who might spin up the bot in their own servers, not just uqcs & testing.

andrewj-brown (changes requested):

uqcsbot/uptime.py
49Indium: Completely agree. Have modified to correct for this.

andrewj-brown (approved):
lgtm

PR #88, opened 2023-04-10
consistent handling of channels using guild lookup

### LINKED ISSUE #87: IDs are silly
We should have a universal system for handling (well, avoiding) IDs in the bot. Possibly an env variable.

andrewj-brown:
fixes #87.

will be merged after membercount (#84), which is to-be-merged using the get_all_channels() system. I've prefilled the prod environment with the relevant env variable.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,8 @@
 # Required variables, ensure these are populated
 DISCORD_BOT_TOKEN=
 POSTGRES_URI_BOT=sqlite:///
+SERVER_ID=
+# uqcs server ID is 813324385179271168; uqcs testing server ID is 836589565237264415
 
 # Alternative variable if you're using Docker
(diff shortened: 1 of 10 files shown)


49Indium:
Do we want checks that the server and channels exist, or is it just good enough to crash and burn at that point.

andrewj-brown:
> Do we want checks that the server and channels exist, or is it just good enough to crash and burn at that point.

If the server doesn't exist, then it's probably ok to crash and burn as a fundamental "the bot has to be in at least one server, and know which server that is" kinda way. However, if the channels don't exist it's probably worth outputting a comprehensible error message so people know which channel(s) their test servers might be missing. I'll investigate.

andrewj-brown:
n.b. haiku doesn't need error message handling; it'll just silently not work if channel attempts return None because it never .sends stuff to them.

andrewj-brown:
there's probably a Real Way to switch which branch this pr takes from but i cba figuring it out, new pr inbound soon

PR #90, opened 2023-04-10
Haiku: Better distinction of words ending in "ate"

49Indium:
Added better determination of syllables in words ending in "ate", alongside relevant tests and a few more exceptions that do not fit into any large groups.
Slight code organisation changes to "le" ending words as well.


andrewj-brown:
the exact vagaries of the haiku input-output are a mystery to me, so if you reckon it's an improvement, I can certify it won't blow up the bot, lgtm

PR #91, opened 2023-04-10
Support non-ASCII character encodings in binify

TheOnlyMrCat:
Binify is currently limited to ASCII text. ASCII was a standard that originated in 1963, for use on old teleprinters. I think we can do better than that.

This patch extends the binify command to work with every character coding supported by python (a list of which can be found here: https://docs.python.org/3/library/codecs.html#standard-encodings). In addition, it sets the default character encoding to UTF-8, as it should be.

This patch also adds the hexify command, which does exactly the same thing as binify, but in hexadecimal.

### DIFF
--- a/uqcsbot/text.py
+++ b/uqcsbot/text.py
@@ -7,6 +7,14 @@ from discord import app_commands
 from discord.ext import commands
 
 
+async def encoding_autocomplete(interaction: discord.Interaction, current: str) -> list[app_commands.Choice[str]]:
+    encodings = ["utf-8", "utf-16", "utf-32", "latin-1", "iso-8859-1", "ascii", "cp037", "cp437", "utf-7"]
+    return [
+        app_commands.Choice(name=encoding, value=encoding)
+        for encoding in encodings if current.lower() in encoding.lower()
+    ]
+
+
 class Text(commands.Cog):
     def __init__(self, bot: commands.Bot):
         self.bot = bot
@@ -30,6 +38,7 @@ class Text(commands.Cog):
 
     @app_commands.command()
     @app_commands.describe(message="Input string", encoding="Character encoding to use, defaults to UTF-8")
+    @app_commands.autocomplete(encoding=encoding_autocomplete)
     async def binify(self, interaction: discord.Interaction, message: str, encoding: Optional[str] = "utf-8"):
         """
         Converts a binary string to text or vice versa.
@@ -83,6 +92,7 @@ class Text(commands.Cog):
 
     @app_commands.command()
     @app_commands.describe(message="Input string", encoding="Character encoding to use, defaults to UTF-8")
+    @app_commands.autocomplete(encoding=encoding_autocomplete)
     async def hexify(self, interaction: discord.Interaction, message: str, encoding: Optional[str] = "utf-8"):
         """
         Converts a hexadecimal string to text or vice versa.
--- a/uqcsbot/text.py
+++ b/uqcsbot/text.py
@@ -1,4 +1,5 @@
 from random import choice, randrange
+from string import hexdigits
 from typing import Optional
 
 import discord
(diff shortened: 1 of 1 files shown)


uqcsbot/text.py
```diff
@@ -28,33 +29,33 @@ def __init__(self, bot: commands.Bot):
         self.bot.tree.add_command(self.scare_menu)
 
     @app_commands.command()
-    @app_commands.describe(message="Input string")
-    async def binify(self, interaction: discord.Interaction, message: str):
+    @app_commands.describe(message="Input string", encoding="Character encoding to use, defaults to UTF-8")
+    async def binify(self, interaction: discord.Interaction, message: str, encoding: Optional[str] = "utf-8"):
```
49Indium, line 33: Making encoding a `Optional[Literal["utf-8", "ascii"]]` would allow discord to give auto-suggestions for the encoding. `Literal["utf-8", "ascii"]` could also be made its own type.

uqcsbot/text.py
```diff
@@ -76,6 +77,31 @@ async def caesar(self, interaction: discord.Interaction, message: str, distance:
 
         await interaction.response.send_message(result)
 
+    @app_commands.command()
+    @app_commands.describe(message="Input string", encoding="Character encoding to use, defaults to UTF-8")
+    async def hexify(self, interaction: discord.Interaction, message: str, encoding: Optional[str] = "utf-8"):
```
49Indium, line 82: Similarly to above, encoding should be of type `Optional[Literal["utf-8", "ascii"]]` or similar.

uqcsbot/text.py
```diff
@@ -28,33 +29,33 @@ def __init__(self, bot: commands.Bot):
...
-                        break
-                    response += chr(n)
+                decoded_message = bytearray()
+                for i in range(0, len(message), 8):
+                    n = int(message[i:i+8], 2)
+                    decoded_message.append(n)
+                try:
+                    response = decoded_message.decode(encoding)
+                except UnicodeDecodeError as e:
+                    response = e.reason
```
49Indium, line 50: Maybe add a beginning description to the error like `f"Could not convert to text: {e.reason}"`, but this is just my personal opinion. As long as you're consistent it's alright.

49Indium (changes requested):
Looks great. The auto-suggestions would be nice to add as well.

uqcsbot/text.py
TheOnlyMrCat: I think the no-preamble option is slightly more consistent, considering our manual "partial byte" errors also have no preamble. As an example, trying to decode `10001001` (or any single byte starting with `10`) gives the error message "invalid start byte", which I think is clear enough.

uqcsbot/text.py
49Indium: You would want to list all the options from what I've read [here](https://discordpy.readthedocs.io/en/stable/ext/commands/commands.html).

andrewj-brown:
Heya, just a heads-up that we have incoming changes (here)[https://github.com/UQComputingSociety/uqcsbot-discord/pull/94] to our repo's code style that will be royal hell to merge into existing PRs. There's currently no ETA on us merging this; we want to give our extant PRs time to get merged beforehand.

If there's anything I can do to help get your PR merged (testing, reviews, etc) shoot me a ping on discord @</hax>#6701

uqcsbot/text.py
TheOnlyMrCat: That would mean making a `Literal[]` listing every single codec on the page...
Or choosing just a subset of them to list. I don't like either solution particularly much, but I'm happy to do it if it's preferable to having zero autosuggestions.

andrewj-brown: I now *have* looked into this. I think a limited set of autosuggestions is the best case here - my dodgy union type idea doesn't work, and in any case, allowing arbitrary str input can throw `LookupError: unknown encoding`, which breaks the bot.

I reckon just pick your favourite 5-10 English encodings and add them to a literal type.

TheOnlyMrCat: How nice are you imagining the errors? Would a simple link to the list page be ok?

> "Invalid encoding. A list of valid encodings can be found at https://docs.python.org/3/library/codecs.html#standard-encodings"

Otherwise a rough autosuggestion might be good, but too complex to add in this PR.

andrewj-brown: I'm happy with basically anything that's not "The application did not respond" tbh; a link sounds good to me.

TheOnlyMrCat: It looks like autocomplete suggestions can be done with a function decorator as well: https://discordpy.readthedocs.io/en/stable/interactions/api.html#discord.app_commands.autocomplete
Doing that still allows arbitrary string input

andrewj-brown: TIL, and if you want to whack one of them on, be my guest

TheOnlyMrCat: Done!

49Indium: Looks great!

uqcsbot/text.py
49Indium: Same as above. Completed with autocomplete

49Indium (approved):
Looks great

PR #95, opened 2023-04-13
Haiku: Further edge cases

49Indium:
Even more edge cases fixed, plus a bit of an overhaul of the internal logic. All test cases (plus new ones) still pass.


lcox74:
Would it be inappropriate to recommend a natural language package like `NLTK`? They have an unhealthily large pronunciation dictionary in the [corpus](https://www.nltk.org/howto/corpus.html) module. It could be side loaded with your unicode sanitisation.

Otherwise, what you're doing seems like witchcraft to me.

49Indium:
> Would it be inappropriate to recommend a natural language package like `NLTK`? They have an unhealthily large pronunciation dictionary in the [corpus](https://www.nltk.org/howto/corpus.html) module. It could be side loaded with your unicode sanitisation.

Similar approaches have been discussed quite a bit. I am against them for a few reasons. Firstly, dictionary look-ups don't deal well with technical words (which are common in CS) and made up words. They also can be more difficult to add our own exception cases for. Lastly, a more algorithmic approach is much more fun!

With the last few commits making the exception lists greater (though still containing very few words), I could be convinced to change to a dictionary approach.

Also, I'm planning on adding an `explain` option to `/numberofsyllables` after this PR, to explain how the bot deals with any example word. This might help explain some of the internals.

JamesDearlove (approved):
lgtm

PR #96, opened 2023-04-13
xkcd

lcox74:
As I delved into the legendary unimplemented folder, seeking refuge from the monotony of work, my eyes fell upon it `xkcd.py` a forgotten relic that once brought joy to all who used it. The voices in my head erupted in a symphony of sarcastic encouragement:

- `"Python isn't that hard", `
- `"This will only take a couple minutes",`
- `"This surely won't break the second someone uses it"`
- `"What can possibly go wrong"`

Hours flew by as I pounded away at the keys, the infectious beats of the `Back Pocket` stream fueling my determination. And now, behold! I have forged a miracle - a feat of coding prowess that even benefits the maintainers of the repository by giving them more work to do.

Of course, I understand that you have more pressing dev-ey tasks at hand (#94), so feel free to defer review of this pull request.

### DIFF
--- a/uqcsbot/xkcd.py
+++ b/uqcsbot/xkcd.py
@@ -1,0 +1,104 @@
+import requests
+import re
+import html
+
+from typing import Optional
+
+import discord
+from discord import app_commands
+from discord.ext import commands
+
+from uqcsbot.bot import UQCSBot
+
+XKCD_BASE_URL = "https://xkcd.com/"
+
+XKCD_FETCH_ERROR = (-1, "", "", "") # xkcd failed to fetch page
+XKCD_PARSE_ERROR = (-2, "", "", "") # xkcd failed to parse page
+
+class Xkcd(commands.Cog):
+
+    def __init__(self, bot: UQCSBot):
+        self.bot = bot
+
+    @app_commands.command(name="xkcd")
+    @app_commands.describe(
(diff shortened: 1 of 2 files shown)


uqcsbot/xkcd.py
```diff
@@ -0,0 +1,104 @@
...
+        :param url: The url to fetch the xkcd data from
+        :return: A tuple containing the xkcd number, title, description and image url
+        """
+
+        # Get the xkcd page
+        response = requests.get(url)
+        if response.status_code != 200:
+            return XKCD_FETCH_ERROR
+        
+        data = str(response.content)
```
49Indium, line 83: This does not handle unicode correctly (consider comics [1137](https://xkcd.com/1137/), [1913](https://xkcd.com/1913/) and [2694](https://xkcd.com/2694/)).
A potential solution is to use `str(response.content, encoding="utf-8")`. Note that if you do this, the description regex won't work, so you'll also have to do something like:
`desc_search = re.search(r'(?<=<div id="comic">).*?title="(.*?)".*?(?=</div>)', ''.join(data.splitlines()), re.MULTILINE)`

49Indium (changes requested):
Unicode strikes again! This time it should be fixable.

Otherwise looks amazing and should be a lovely addition.

49Indium (approved):
LGTM

PR #97, opened 2023-04-15
Member count

49Indium:
`/membercount` did not work on the main server (despite working on the test server) due to permissions. Here are some error catching and logging mechanisms to allow easy checking of bot permissions and reasons for failure.

### DIFF
--- a/uqcsbot/member_counter.py
+++ b/uqcsbot/member_counter.py
@@ -48,7 +48,7 @@ class MemberCounter(commands.Cog):
             for member in interaction.guild.members
             if member.joined_at > datetime.now(tz=ZoneInfo("Australia/Brisbane")) - self.NEW_MEMBER_TIME
         ]
-        await interaction.response.send_message(f"There are currently {interaction.guild.member_count} members in the UQCS discord server, with {len(new_members)} joining in the last 7 days.")
+        await interaction.response.send_message(f"There are currently {interaction.guild.member_count} members in the UQ Computing Society discord, with {len(new_members)} joining in the last 7 days.")
 
     @commands.Cog.listener()
     async def on_member_join(self, _):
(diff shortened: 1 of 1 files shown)


uqcsbot/member_counter.py
```diff
@@ -37,16 +38,22 @@ async def on_ready(self):
...
-            lambda member: member.joined_at > datetime.now() - self.NEW_MEMBER_TIME,
-            interaction.guild.members
-        )
-        await interaction.response.send_message(f"There are currently {interaction.guild.member_count} members in the UQCS discord server, with {len([new_members])} joining in the last 7 days.")
+        new_members = [
+            member
+            for member in interaction.guild.members
+            if member.joined_at > datetime.now(tz=ZoneInfo("Australia/Brisbane")) - self.NEW_MEMBER_TIME
+        ]
+        await interaction.response.send_message(f"There are currently {interaction.guild.member_count} members in the UQ Computing Society discord, with {len(new_members)} joining in the last 7
```
andrewj-brown, line 56: This is broken...

andrewj-brown (changes requested):
Dunno what happened here, but it's not compiling on my machine.

I incorporated a membercount --force argument in the #94 pr, which saves needing to restart the bot if it gets stuck. The new error handling should stop those situations from occurring in the first place, but it may still be useful to have a "force exit blocked state" button somewhere, just in case.

uqcsbot/member_counter.py
49Indium: Oops. Merging with main got rid of the end of that that line by mistake. Fixed by appending ` days")`. Should be fixed in latest commit.

49Indium:
> I incorporated a membercount --force argument in the #94 pr, which saves needing to restart the bot if it gets stuck. The new error handling should stop those situations from occurring in the first place, but it may still be useful to have a "force exit blocked state" button somewhere, just in case.

I still think that this is a great option, so I'm looking forward when this is merged. I didn't want to recode that, so my intention was to wait until that was completed.

andrewj-brown:
> I still think that this is a great option, so I'm looking forward when this is merged. I didn't want to recode that, so my intention was to wait until that was completed.

#94 is going to be blocked for quite some time while I deal with uni, and in the meantime we're still unable to debug membercount without restarting. I reckon just add it to this PR and I'll deal with #94 at a later date.

andrewj-brown (approved):
lgtm

PR #111, opened 2023-05-05
Update workflow and deps

### LINKED ISSUE #105: kill joinchannel cog
Putting this up as a reminder that when I'm not snowed in by assignments, joinchannel needs to be deleted from the bot.

Quite possibly this will also need manual fiddling to remove the old channel tables from the DB, hence committee-only label.

JamesDearlove:
Updating workflow dependencies, and also package dependencies as well.

While I was here, I removed the channel cog and dealt with some quick deprecation warnings (ie SQLAlchemy 2)

Closes #105 

**Update 10th May** I've also now gone and:
- Added a Pytest workflow to run Pytest on PRs
- Removed Docker build workflow from PRs
- Rewritten Dockerfile to be multi-stage, which should speed up deployment
- Fixed @andrewj-brown's merge issue


andrewj-brown (approved):
LGTM conditional on the #94 changes being rebased in.

I assume this'll fix all our deprecation warnings? I can't see anything that touches `save-state` or `set-output`, unless they're pulled in via the deps you've updated.

JamesDearlove:
> I assume this'll fix all our deprecation warnings? I can't see anything that touches `save-state` or `set-output`, unless they're pulled in via the deps you've updated.

Yeah all those seemingly were part of the outdated deps, which is very convenient

README.md
```diff
@@ -64,6 +64,14 @@ To run a particular test, say `test_whatweekisit.py`, run:
 poetry run pytest tests\test_whatweekisit.py
 ```
 
+## Code Styling
+
+We use an automated code formatter called [Black](https://black.readthedocs.io/), currently this needs to be run manually to pass the format CI check. To run Black, run from the root of the repo:
+
+```bash
+poetry run black uqcsbot
```
andrewj-brown, line 72: does that work? poetry just inherently knows which files to do? massive

andrewj-brown (approved):
lgtm still

README.md
JamesDearlove: Well in this case, its running it on the folder uqcsbot so it's smart enough to know what files to run it on otherwise

PR #113, opened 2023-05-07
remove /echo (pt 2, this time, not based off starboard branch 💀)

### LINKED ISSUE #107: kill /echo
/echo can be used for a variety of things. Lots of them are bad.

Sanitising it properly (i.e. removing all the possible "Bad Things" that can be echo'ed) is far too much work, and let's be honest, what's the real need for it anyway.

Therefore, we should just kill it. Timeline tbd.

andrewj-brown:
Pretty self-explanatory. Remains necessary while Discord automod doesn't flag bots - /echo is basically just a free automod-bypass, as well as being able to mimic system messages and the outputs of other commands.

https://support.discord.com/hc/en-us/articles/4421269296535-AutoMod-FAQ
> Q: Will AutoMod flag messages from bots?
> A: No - all bot alerts and messages will be exempt from AutoMod.

Closes #107.


JamesDearlove (approved):

andrewj-brown:
It should go without saying, but if any discord members see this PR and decide to "test out" the automod bypass, committee will be *extremely* less than amused.

Please don't. They didn't sign up because they wanted to moderate a discord server.

Issue #56, opened 2022-11-13
Basic type-checking and annotations


katrinafyi:
It would be nice to have some type-checking (like mypy) which runs on the CI build. This would help catch bugs such as those in #54 and is easier than comprehensive unit testing.

I have vague memories of [UQComputingSociety/uqcsbot-slack](https://github.com/UQComputingSociety/uqcsbot-slack) having this, though I might be mixing it up with ordinary unit tests (which are also not on Discord, but that's #2).

I would do this myself but I suspect the bot is currently in a state of discord.py limbo. Maybe whenever that is sorted out.

JamesDearlove:
Hoping to have discord.py limbo sorted soon™ @kumar-ish mentioned this in his PR and it's a good idea to have something like this

PR #114, opened 2023-05-07
add static typecheck system to poetry

andrewj-brown:
From one wide-ranging-dev-changes pull request to the next - now we have code style, I'm starting work on issue #56.

First step: get mypy to not throw errors on the current code, then I'll add a GH workflow for it for future branches. I've based this off the #111 PR branch, because I don't want to bother typing the entire `channels` cog when it's about to be removed.


andrewj-brown:
once more, this time with feeling...

49Indium:
Do we want to use `list`, `tuple`, etc instead of `List`, `Tuple`, etc. I only learnt recently that are already included (no need for Typing) from Python 3.9 onwards,

andrewj-brown:
I'm in favour of standardising and using the `typing` types for everything, rather than mixing and using some inbuilt types and some `typing` types.

andrewj-brown:
I intend to get this PR merged as-is. The changes so far allow use of `poetry run pyright uqcsbot` to statically typecheck commands, or `... pyright uqcsbot/file.py` for specific files.

My intention is to require that pyright types all new modules from here on out, including the proposed `pastexams`, `dominos`, `morse`, etc. rather than merge them now and just leave the typing until later.

uqcsbot/models.py
```diff
@@ -1,56 +1,56 @@
...
+    id: Mapped[int] = mapped_column("id", BigInteger, primary_key=True, nullable=False)
+    user_id: Mapped[int] = mapped_column("user_id", BigInteger, nullable=False)
+    channel_id: Mapped[Optional[int]] = mapped_column(
+        "channel_id", BigInteger, nullable=True
+    )
+    time_created: Mapped[int] = mapped_column("time_created", DateTime, nullable=False)
+    message: Mapped[str] = mapped_column("message", String, nullable=False)
+    time = mapped_column("time", Time, nullable=False)
+    start_date = mapped_column("start_date", Date, nullable=False)
+    end_date = mapped_column("end_date", Date, nullable=True)
```
49Indium, line 50: I'm unsure of protocol: do these need types?

uqcsbot/models.py
```diff
@@ -1,56 +1,56 @@
...
-    channel_id = Column("channel_id", BigInteger, nullable=True)
-    time_created = Column("time_created", DateTime, nullable=False)
-    message = Column("message", String, nullable=False)
-    time = Column("time", Time, nullable=False)
-    start_date = Column("start_date", Date, nullable=False)
-    end_date = Column("end_date", Date, nullable=True)
-    week_frequency = Column("week_frequency", Integer, nullable=True)
+    id: Mapped[int] = mapped_column("id", BigInteger, primary_key=True, nullable=False)
+    user_id: Mapped[int] = mapped_column("user_id", BigInteger, nullable=False)
+    channel_id: Mapped[Optional[int]] = mapped_column(
```
49Indium, line 43: I'm unsure why this one is `Mapped[Optional[int]]`. I don't think it needs a comment, just personal curiosity, as I'm not that familiar with this section of the code.

49Indium (commented):
Take everything I say with a pinch of salt, as I'm still learning good coding practices. Here are my 2 cents.

uqcsbot/models.py
andrewj-brown: I think they will in order to be used properly in other files, but the other files are excluded for now until I get the full PR done.

uqcsbot/bot.py
```diff
@@ -35,9 +35,9 @@ async def admin_alert(
         self,
         title: str,
         colour: discord.Colour,
-        description: str = None,
-        footer: str = None,
-        fields: List[tuple] = None,
+        description: Optional[str] = None,
+        footer: Optional[str] = None,
+        fields: Optional[List[tuple]] = None,
```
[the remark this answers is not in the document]
andrewj-brown: `bot.py` doesn't typecheck yet, it's still excluded. You're probably right, but then it'll also need to be `Tuple[something else]` as well.

uqcsbot/models.py
andrewj-brown: Hm, I'm pretty sure `nullable=True` is what needs `Optional`, but that doesn't fully line up with the actual types. My local install is complaining about unknown imports, I'll see if I can get that fixed before we merge.

(weird, pyright still types `models.py` correctly...)

andrewj-brown: Note for posterity: my local typing issues were caused by dependency hell, now fixed.

uqcsbot/advent.py
```diff
@@ -542,5 +546,13 @@ async def advent_winners(
 
 
 async def setup(bot: UQCSBot):
+    if os.environ.get("AOC_SESSION_ID") is not None:
+        SESSION_ID = os.environ.get("AOC_SESSION_ID")
```
JamesDearlove, line 550: I'd rather this to happen within the `Advent` class imo, but I'm not toooo fussed as this cog is on the rewrite chopping block.

JamesDearlove (changes requested):
Seems mostly fine, it appears that your `poetry.lock` needs to be updated though

andrewj-brown:
> Seems mostly fine, it appears that your `poetry.lock` needs to be updated though
This PR intentionally updates `poetry.lock` - we've got 5 new modules for typing and stubs, so I took the opportunity to bump versions on a bunch of other libraries so dependabot stops whinging.

JamesDearlove (approved):
Code looks good to me. Still have a lock file warning in the checks though. [Action log link](https://github.com/UQComputingSociety/uqcsbot-discord/actions/runs/5409447098/jobs/9829600441#step:5:17)

Issue #103, opened 2023-04-28
/morse


49Indium:
We have a lot of text modification commands, but no commands to convert to morse and back. I think this would be a good addition to the bot. Inspired by [this](https://discord.com/channels/813324385179271168/813325735620116490/1101281627902062722).

jenseni-git:
I might jump on this to get used to playing around with the bot, and then try the announcement sender

PR #75, opened 2023-03-20
Starboard 2.0

andrewj-brown:
Starboard 2.0, implementing a bunch of stuff I originally ceebs'ed, plus novel features that committee discovered a need for:
- Starboard reaction totals are now calculated as (reacts on original message + reacts on starboard message)
- These reacts are now enforced one-per-person *and* don't include the author of the original message
- Allowed starboarding messages from admin channels (comedy inbound!)
- Subsequently, changed the message format so that channels display the name, rather than a link (which will be invisible for admin/private channels)
- Committee now have the ability to "blacklist" a message from being starboarded
- Committee now have a "cleanup" command that should remove any race-condition-caused duplicate messages
- Deletions of bot-posted-starboard-messages no longer cause the bot to shit itself
- Misc bugfixes (timezones; ratelimit in .env.example; replies to deleted messages; empty messages)

Many of these features are only minimally tested. Prior to merging I'll need to get a few other people together in the testing server and make sure it's all fully functional.

PR #117, opened 2023-05-08
Add morse command

jenseni-git:
Adds command to convert text to morse code (issue #103).


.gitignore
```diff
@@ -131,3 +131,4 @@ dmypy.json
 .vim
 *.db
 **/.DS_STORE
+.vscode/settings.json
```
andrewj-brown, line 134: similarly-ish, just ignore .vscode/* - it'll save hassle in future if someone else using vsc has to do it

.env.example
```diff
@@ -17,4 +17,4 @@ SB_BIG_THRESHOLD=24
 <<<<<<< HEAD
 SB_RATELIMIT=30
 =======
->>>>>>> 6ce7152bcf2777976527b540ea014b9967d772fa
+>>>>>>> 6ce7152bcf2777976527b540ea014b9967d772fa
```
andrewj-brown, line 20: yeah that's on me, git blame points at #75.

If you could just remove the merge artifacts altogether, that would be big

unimplemented/morse.py
```diff
@@ -0,0 +1,159 @@
+from typing import Optional, Literal, List
```
andrewj-brown, line 1: probably should be removing it from ./unimplemented if morse is done now

uqcsbot/morse.py
```diff
@@ -0,0 +1,159 @@
...
+    ":": "_ _ _ . . .",
+    "'": ". _ _ _ _ . ",
+    '"': ". _ . . _ . ",
+    "_": ". . _ _ . _ ",
+    "=": "_ . . . _ ",
+    "+": ". _ . _ . ",
+}
+
+
+class morse(commands.Cog):
```
andrewj-brown, line 67: Convention/nitpick: cog names are `UpperCamelCase`.

uqcsbot/morse.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        Replaces all illegal characters in the message with their sanitised
+        equivalent.
+        """
+
+        # Strip whitespace from either side of the message
+        message = message.strip()
+
+        # replace code blocks with their sanitised equivalent
+        message = message.replace("```", "'''")
```
andrewj-brown, line 118: Haven't tested - is it worth just stripping all ` characters? If code blocks have issues, inline code will probably still have issues.

andrewj-brown (changes requested):
A few points, mostly just nitpicks/git stuff, nothing wrong with the actual code.

.gitignore
jenseni-git: Done

.env.example
jenseni-git: Removed.

unimplemented/morse.py
jenseni-git: Removed rom ./unimplemented

uqcsbot/morse.py
jenseni-git: Done

uqcsbot/morse.py
jenseni-git: Just copied it from cowsay. Didn't actually read it. If a user prints something that's code, they probably just want the morse of the actual code, so I've replaced "`" with "".

jenseni-git:
Sorry. I now realise re-request review is probably the equivalent of an obnxious ping.

uqcsbot/morse.py
andrewj-brown: It wouldn't be unwise to move "sanitisation" like this out to a set of functions that can be used bot-wide, instead of re-implementing them every cog, but that's out-of-scope here. Sounds good to me :+1:

andrewj-brown:
> Sorry. I now realise re-request review is probably the equivalent of an obnxious ping.

I have every email turned on for this repo. doing *anything* is the equivalent of an obnoxious ping 😂

uqcsbot/morse.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # Replace each emote with its name.
+        for emote in emotes:
+            emote_name: str = emote.split(":")[1].strip()
+            message = message.replace(emote, f":{emote_name}:")
+
+        return message
+
+    @staticmethod
+    def check(message):
```
andrewj-brown, line 140: Rather than continuing to go over the list, you can just instantly return some boolean value for "bad", and then check that directly in the main command (instead of checking list lengths).

We're not massively concerned about performance, but it's still unnecessary effort.

.env.example
```diff
@@ -15,3 +15,4 @@ MC_RCON_PASSWORD=
 SB_BASE_THRESHOLD=8
 SB_BIG_THRESHOLD=24
 SB_RATELIMIT=30
+SB_RATELIMIT=30
```
andrewj-brown, line 18: double env :thonk:

andrewj-brown (changes requested):
I'm happy to merge after these get fixed. Apologies for the long-ass turnaround on this one, it'll be good to close it off soon.

uqcsbot/__main__.py
```diff
@@ -55,6 +55,7 @@ async def main():
         "jobs_bulletin",
         "latex",
         "member_counter",
+        "morse",
```
49Indium, line 58: This is a very small nitpick, but could this be moved a line lower to keep it in alphabetical order.

uqcsbot/morse.py
49Indium, line 140: It would be nice to have doc strings or return types (given by ` -> int` etc) for some of these methods. The name `check` doesn't provide the whole picture.

uqcsbot/morse.py
```diff
@@ -0,0 +1,159 @@
...
+        if invalid != 0:
+            await interaction.response.send_message(
+                f"Invalid morse code character/s in string: {invalid}"
+            )
+            return
+
+        # encrypt message
+        response = morse.encrypt_to_morse(message)
+
+        # Check if response is too long
```
49Indium, line 100: I like this check

uqcsbot/morse.py
49Indium: But it then provides all the "bad" characters to the user, instead of just the first one. I'm not completely against what is already here.

uqcsbot/morse.py
```diff
@@ -0,0 +1,159 @@
...
+
+    @staticmethod
+    def check(message):
+        ret = ""
+        for letter in message:
+            if letter not in MorseCodeDict:
+                ret += letter
+        if len(ret) != 0:
+            return ret
+        return 0
```
49Indium, line 147: Maybe return an empty string if there are no "bad" characters. It simplifies these last few lines and makes the return type of the function more uniform.

49Indium (changes requested):
The main command function and `encode` function are very readable. I've nit-picked about `check()`, but otherwise looks good.

uqcsbot/morse.py
andrewj-brown: ...eh, I guess that's helpful, but I reckon it'd be pretty hard to submit non-encodable characters without realising.

jenseni-git: Yeah I was thinking more along the lines of Indium when I made this. I figured it's more useful for the user to know all of the characters than just lmao ur bad. Especially since some of your general special chars like #$% etc. are not encoded.

49Indium (approved):
LGTM. I'll wait a little bit for any other opinions, otherwise I'll merge tonight.

@andrewj-brown Are you happy with the current changes?

andrewj-brown (approved):
lgtm and merging

PR #122, opened 2023-05-17
Detect if edits are made in #yelling

JamesDearlove:
That's right, we're coming for you. This should now detect if anyone edits their message, but only if they now contain lowercase.

Also, bug fix for daily holiday message not being in the Brisbane timezone


49Indium (commented):
Looks good to me. Personally, `check_lowercase` feels a little less descriptive than `any(char.islower() for char in message)` but does remove duplication, so I'm not opposed to it. Maybe `contains_lowercase()` would be a more appropriate name.

PR #124, opened 2023-06-23
Fixed bug where workingon calls the same person twice

49Indium:
The daily "Tell us about something cool you are working on!" message calls on two people, but could call on the same person twice (as seen [here](https://discord.com/channels/813324385179271168/813325735620116490/1119884203241259128)). This should hopefully stop this rare bug from happening.

### DIFF
--- a/uqcsbot/whatsdue.py
+++ b/uqcsbot/whatsdue.py
@@ -74,18 +74,25 @@ class WhatsDue(commands.Cog):
         except (CourseNotFoundException, ProfileNotFoundException) as e:
             await interaction.edit_original_response(content=e.message)
             return
+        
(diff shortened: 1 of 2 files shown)


andrewj-brown:
lgtm, but only if we're sure it's a bug worth fixing - there's a 1/1444 chance and shrinking, and it gives us something to talk about in the (extremely usual) case that the actual requestee doesn't respond.

49Indium:
Fair enough. I'll let others chime in.

(Re revert: Oops, pushed to wrong branch)

lcox74:
Lgtm. But you're removing character from the bot

bradleysigma:
Yeah, with 1444 members (and counting), you'd expect to go four years between double pings. The change I'd prefer is bumping it up to three pings every day.

Issue #93, opened 2023-04-13
/syllable Miscounting Errors


49Indium:
This issue is for any errors with /syllable miscounting syllables. Due to the large number of words within English, this issue may stick around. If you find a word (or haiku) that uqcsbot gets wrong, please add it here. Words (or haikus) added here should be added to the test cases and hopefully fixed.

A few important points:
- British pronunciation will be used as the guideline of the true number of syllables. This issue is not for debaing the number of syllables in certain words, but to identify words which uqcsbot gets wrong.
- The aim is to avoid dictionary-style searches and large exception lists. These are much less effective at handling technical and made-up words. In doing so, try to identify a root reason (usually a suffix, prefix or root word) why the bot fails with your certain word, so that all words of this case can be fixed at once. (Also, dictionary look-ups aren't as fun)

49Indium:
As discussed on the discord, numbers need better treatment (currently they count as 0 syllables). There were a few options that were brought up:
- Write each digit with its word equivalent: "2020" -> "two zero two zero"
- Replace numbers between 0-99 with its word equivalent and ignore all others
- Replace all numbers with its word equivalent: "2020" -> "two thousand and twenty"
- Replace numbers between 0-99 with its word equivalent, have some cases for years (e.g. "2020" -> "twenty twenty") and ignore all others

Any suggestions on what path to take would be appreciated.

49Indium:
I'm going to close this for now, as no miscounts remain that have been reported, and none have been reported within the last 3 months.

PR #127, opened 2023-06-25
Made haiku detection non-deterministic

49Indium:
Too many messages were being flagged as haiku's, so this makes the bot give a probability to each message on how strongly it believes a message to be a haiku. Punctuation and certain words can then make it more likely or less likely for the bot to respond.

Additionally, the syllable rules have been put into a YAML file so that formatting isn't an issue and to make the code a little less messy. A bunch of exceptions that have been found have also been fixed. This isn't a full fix of #93, as there are still some to go, but a partial improvement.


uqcsbot/haiku.py
```diff
@@ -1,55 +1,103 @@
...
-        "general",
-        "memes",
-        "yelling",
-    ]
-    YELLING_CHANNEL_NAME = "yelling"
-
     def __init__(self, bot: UQCSBot):
         self.bot = bot
+        if affixes == {} or syllable_exceptions == {} or accent_replacements == {}:
+            raise FatalErrorWithLog(
```
andrewj-brown, line 71: fwiw `FatalErrorWithLog` isn't designed for `__init__` contexts, as it needs the admin-alerts channel (which isn't available until ready).

Also, generally, errors in `__init__` context are fatal, because discord.py hasn't setup its callback failure recovery systems around it. Raising a `FatalErrorWithLog` here is effectively the same as a `RuntimeError` in terms of output.

uqcsbot/haiku.py
```diff
@@ -74,350 +122,99 @@ async def syllables(self, interaction: discord.Interaction, word: str):
...
+    probability: float = HAIKU_BASE_PROBABILITY
+    syllable_count: int = 0
+    lines: List[str] = []
+    current_line: List[str] = []
+    punctuation_at_end_of_previous_line = True  # Initially true so that any punctuation at the beginning does not increase the probability.
     haiku_syllable_count = [5, 7, 5]
+
+    def _increased_probability(probability: float, index: float):
+        """
+        Calculates the probability of at least 1 success in index Bernoulli trials with given probability.
```
andrewj-brown, line 139: statistician moment :heart:

uqcsbot/haiku.py
andrewj-brown: On a related note, there's a bit of a mix of errors going on here - due to the "global" (i.e. outside Haiku class) handling of the file stuff, if the file is straight up missing the the bot will also refuse to start (`FileNotFoundError`), but if the file is there and just incomplete the bot will refuse to start (`FatalErrorWithLog`) from a completely different context & with a different exception.

I'd recommend replacing this with a completely new system that just throws all the "can't figure out what's going on with syllable rules" exceptions from one spot. Discussion to happen on discord about `on_ready` vs `__init__` shortly, because I believe jimmy has opinions.

andrewj-brown (changes requested):
the haiku stuff: lgtm. the error stuff: not so much. Pending discussion.

andrewj-brown (approved):
lgtm and merging

PR #133, opened 2023-06-30
use cache for both poetry workflow runs

andrewj-brown:
by caching poetry dependencies, we should hopefully significantly speed up the testing and the typing runs.

Theoretically the "fastest" option is to install pyright/pytest on their own, but then we have two places where dependencies are stored, instead of the single-source-of-truth being poetry.


andrewj-brown:
...first cache hit, looks like the net saving is ~5 seconds? Poetry accounts for the *vast* majority of that setup time

JamesDearlove (changes requested):
I hate to be that person, but there's a much simpler way: https://github.com/actions/setup-python#caching-packages-dependencies

Imo if you're looking for improving testing run times, I think that combining the static check + unit test actions would be better. After all, they are both essentially testing the code itself. Black can remain separate, as it can run in parallel to these and doesn't require poetry.

andrewj-brown:
> I hate to be that person, but there's a much simpler way: https://github.com/actions/setup-python#caching-packages-dependencies
> 
> Imo if you're looking for improving testing run times, I think that combining the static check + unit test actions would be better. After all, they are both essentially testing the code itself. Black can remain separate, as it can run in parallel to these and doesn't require poetry.

There was smething about unable to pass certain parameters into poetry when using setup-python's cache, but I'll give it a look. If we use setup-python's cache it might even cache the actual poetry install, which would be significantly faster.

andrewj-brown:
Yeah OK so I can make it cache the actual poetry install, which will be the fastest option (saves ~19s of poetry install on cache hit) but completely locks us out of using setup-python's autocaching because it requires us actually caching the .poetry folder... which is the opposite of your suggestion :skull:

I'd prefer to keep tests & typing separate so its easier to see what actually failed on a particular PR, although we don't really have enough tests for that to be a massive problem (yet).

andrewj-brown:
Alright, we're now successfully caching poetry *and* the poetry dependencies.

It looks like the `test` workflow has gone from ~40s to ~20s, and the `typecheck` workflow has gone from ~40s to ~30s. (I'm unable to work out how the heck those numbers work, given we've eliminated the same steps from both workflows, but whatever.)

I don't believe it's possible to cache anything else, or if its even worth it considering the setup is now 13s of basically just "download the cache files".

JamesDearlove (approved):
Yeah look, I don't think you're gonna improve much more with the action run times here. I think that even below a minute is still very solid.

Still though, good work

PR #135, opened 2023-07-02
Update to /repos command group

49Indium:
A few changes to the `/repos` command group:
- Changed `list` and `find` to run as embeds
- Added issues, which lists the issues of a specific repo
- Added prs, which lists the PRs of a specific repo

This will require a new auth token in the bot for github API access, which is given by an environment variable. I will update the wiki soon.


andrewj-brown (approved):
LGTM. Awaiting committee discussion on setting up a token.

Be aware that this is a Recurring Task:tm: that will need to be managed by future committees. Github tokens are inherently temporary; there will come a time when the token dies and whoever is around then will need to know what's going on.

andrewj-brown:
Also converted to draft while we figure out what to do with tokens.

PR #136, opened 2023-07-03
Dominos: Neatened embed

49Indium:
The embed would show an empty "Keywords: " line even if there were no keywords given. Also italicized the keywords (if given).


andrewj-brown (approved):
lgtm and merging

Issue #139, opened 2023-07-22
LaTeX command does not accept dollar signs


49Indium:
The LaTeX command should probably not render nothing when dollar signs are in the input argument. I would suggest checking for dollar signs then if they are in the input (and not escaped with a `\`) then start LaTeX generation in horizontal mode (rather than math mode).

An example message can be seen [here](https://discord.com/channels/813324385179271168/836243768411160606/1132129590861639760).

PR #141, opened 2023-07-22
LaTeX: Added error checking for the equation

49Indium:
A small fix to catch when the equations cannot be rendered. Does slow down `/latex` slightly, as image is requested once by the bot and once by discord.

This partially solves #139, though we might need to consider running LaTeX on the bot's side in future (codecogs is a pretty average emulation of TeX IMO). I'd like to discuss the security and costs of this.


andrewj-brown (approved):
LGTM, although this is a small pr, in a fairly disused cog, with future changes planned. It might be prudent to wrap this up with the rest of the LaTeX changes needed to resolve #139.

PR #142, opened 2023-07-28
Whatsdue: Added Sorting

49Indium:
As per [here](https://discord.com/channels/813324385179271168/813404583103168594/1133881745926787153), added sorting for whatsdue. Can now sort by due date, weight and course name.

Also neatened up `uq_course_utils` by adding an Assessment dataclass.


49Indium:
Please don't merge yet, as I'll add the other features wanted by #145 (hopefully this week).

uqcsbot/utils/uq_course_utils.py
```diff
@@ -105,6 +106,72 @@ def _estimate_current_semester() -> SemesterType:
...
+    def is_after(self, cutoff: datetime):
+        """
+        Returns whether the assessment occurs after the given cutoff.
+        """
+        try:
+            start_datetime, end_datetime = self.get_parsed_due_date()
+        except DateSyntaxException:
+            # TODO bot.logger.error(e.message)
+            # If we can't parse a date, we're better off keeping it just in case.
+            # TODO(mitch): Keep track of these instances to attempt to accurately
```
andrewj-brown, line 146: Noting that these TODO's are 2 years old. At a glance, I can't tell how simple these are to fix - it looks like it's failing 'safely', so maybe just delete the TODO's?

uqcsbot/utils/uq_course_utils.py
```diff
@@ -105,6 +106,72 @@ def _estimate_current_semester() -> SemesterType:
...
+            start_datetime, _ = self.get_parsed_due_date()
+        except DateSyntaxException:
+            # TODO bot.logger.error(e.message)
+            # If we can't parse a date, we're better off keeping it just in case.
+            # TODO(mitch): Keep track of these instances to attempt to accurately
+            # parse them in future. Will require manual detection + parsing.
+            return True
+        return start_datetime <= cutoff
+
+    def get_weight_as_int(self):
```
andrewj-brown, line 165: The lack of explicit optionality in types makes me nervous here, considering the name implies it's always an int. Is it worth returning 0 on failure here and then removing the lambda-or thing elsewhere?

andrewj-brown (changes requested):
LGTM (almost). Two very minor things. 

Happy to be wontfix'ed on both, just making sure they're considered.

uqcsbot/utils/uq_course_utils.py
49Indium: I agree with removing them. I tried to find cases where it failed to parse, and where it did, it was because there was no good date to use.

uqcsbot/utils/uq_course_utils.py
49Indium: Potentially this might lead to uglier code in the future, but definitely cleans up code as is, so I'm happy to change.

andrewj-brown (approved):
LGTM and merging

PR #152, opened 2023-08-25
Timeout Members Who Don't Yell In #yelling

bradleysigma:
With this change, when somebody fails to yell in yelling they'll be timed out for 15 seconds. If they do so again, they'll be timed out for 30 seconds, then a minute, then two, and so on. Every day, one violation for each member will be cleared.


49Indium (approved):
Looks great

JamesDearlove (approved):
what a time to be alive

PR #153, opened 2023-08-26
Timeout on edits in #yelling too.

bradleysigma:
Now editing messages in #yelling will result in a timeout if the message isn't yelled.


49Indium (approved):
LGTM

JamesDearlove (approved):
sure why not

PR #155, opened 2023-08-26
 Create Decorator to Prevent Use of Commands With Minuscule Argument To Bot Commands In #yelling

bradleysigma:
The decorator `@yelling_exemptor()` applied to a command will now prevent that command from being used in `#yelling`. Argument can be given to the decorator, which will apply the minuscule filter to those command arguments e.g. `input_args=["message"]` will prevent the command from being used for a command that has `message` as an argument.


49Indium (approved):
Looks great

PR #156, opened 2023-08-27
Trigger Yelling Filter On Bad URLs

bradleysigma:
UQCSbot will try to access any URLs posted in #yelling. If they're invalid, it will trigger the yelling response.


LimaoC (approved):

JamesDearlove (approved):
I think this should be probably fine, uh assuming we don't ddos someone on accident

PR #157, opened 2023-08-27
Better Detection of Emoji in #yelling

bradleysigma:
Now properly detects fake emoji as yelling violations and animated emoji as non-violations.


LimaoC (approved):
yeah nah looks about right

uqcsbot/yelling.py
```diff
@@ -154,16 +156,21 @@ def clean_text(self, message: str) -> str:
 
         # ignore emoji and links
         text = re.sub(
-            r":[\w\-\+\~]+:",
+            r"<(?P<animated>a?):(?P<name>[a-zA-Z0-9_]{2,32}):(?P<id>[0-9]{18,22})>",
```
andrewj-brown, line 159: I'd prefer this to use `\w` and `\d` instead of the character classes `[a-zA-Z0-9_]` and `[0-9]`.

andrewj-brown (changes requested):
regex whinging:tm:

uqcsbot/yelling.py
bradleysigma: I just copied the regex from an admin's message in the discordy.py server ¯\_(ツ)_/¯. Anyway, fixed.

andrewj-brown (commented):
lgtm

andrewj-brown (approved):
lgtm 2 (remember to use the radio buttons this time)

PR #161, opened 2023-08-31
Update yelling.py

bradleysigma:
Handle committee not yelling in #yelling  
* Gives a `:coup:` response if a member fails to be timed out.
    * Also prevents an exception from being logged.
* Doesn't decrement a committee member's counter each day, since they didn't serve their time.

### DIFF
--- a/uqcsbot/yelling.py
+++ b/uqcsbot/yelling.py
@@ -49,7 +49,9 @@ def yelling_exemptor(input_args: List[str] = ["text"]) -> Callable[..., Any]:
                 return
             await interaction.response.send_message(str(discord.utils.get(bot.emojis, name="disapproval") or ""))  # type: ignore
             if isinstance(interaction.user, discord.Member):
-                await Yelling.external_handle_bans(bot, interaction.user)  # type: ignore
+                banned = await Yelling.external_handle_bans(bot, interaction.user)  # type: ignore
+                if not banned:
+                    await interaction.response.send_message(str(discord.utils.get(bot.emojis, name="coup") or ""))  # type: ignore
 
         return wrapper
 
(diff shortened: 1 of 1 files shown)


bradleysigma:
NB: This was all done freehand, with no testing.

JamesDearlove:
Honestly, I'm not convinced these are changes that should be made.

The second part where committee incidents don't decrement also feels like an active attack once a committee member finishes their term. We shouldn't be bullying committee for being moderators of the Discord and therefore incidentally exempt from yelling "rules".

My stance is PR denied, but @UQComputingSociety/committee should make the final decision here.

ew-b:
I don’t fully understand what this does but from what I gather I’m against it. As much as it’s a bit it’s important for committee to be able to effectively communicate and moderate the discord, yes that means there are discord permissions committee has over members but that how it goes. In times where seriousness is needed I think this would hinder it. Committee isn’t exempt from the rest of bot yelling, thus I don’t feel this is appropriate

andrewj-brown (changes requested):
Completely unrelated to committee discussion - there shouldn't be any need for `#type: ignore` comments in this PR.

You can get the emoji and cast it to a non-optional type by saving it in an on_ready handler, which I strongly prefer as it makes it more obvious what emojis the bot will *require* to function normally on a testing server.

bradleysigma:
Counterargument:
Screen Shot 2023-08-31 at 20 39 47

jenseni-git:
> Counterargument: Screen Shot 2023-08-31 at 20 39 47

Counter counter argument

Fair cop, and I'll admit a slight abuse of power. But, with context, that was responding to an @/committee ping.

I think the negative consequences of this PR outway the benefits. There may be a need for testing of things and/or moderation in #yelling, and imo the scale of this change, without necessary balances to cancel possible genuine needs of non-shouting, prevents this from getting my approval.

jenseni-git:
I'm not super against the bot reacting :coup:, but at the same time, I feel like if there's a genuine need to be moderating, between either having to yell or the bot being funny, the message we try to portray will not get across.

andrewj-brown:
Also I slept on it and noticed there's also `#type: ignore` introduced by PR #155, so consider this a retro-active "fix that one too please" considering I wasn't able to review that one before merge.

That one will probably involve creating a new Cog subclass to explicitly codify that our cogs all save the bot to `self.bot`, which - again - is better than just ignoring types, because now we're making that assumption (that cogs all have `self.bot`) explicit in the system.

andrewj-brown:
In #162 I've removed the extant `#type: ignore`'s, however, that only fixes the current state of the bot; I still won't merge this PR while there are `#type: ignore`'s in it.

Issue #160, opened 2023-08-31
Commands do not yell in #yelling 


49Indium:
Commands such as /weekisit do not yell in #yelling.

AncientMeme:
I've decided to comb through the code for a bit. I discovered that commands that output texts calls `yelling_exemptor` from yelling.py. I wonder if that's blocking the bot from yelling out commands.

49Indium:
Thanks for looking into this! I don't think that this is the issue, as `yelling_exemptor` just allows you to run commands with lowercase arguments. For example, `/whatsdue` has `yelling_exemptor`, so you don't get blocked when you search for whats due in `csse1001` (note the lowercase). In contrast, `/cowsay` does not have the `yelling_exemptor`, so trying to run `/cowsay hello` will get blocked by the bot in #yelling (try it for yourself).

I don't blame you for being confused. I think the documentation for `yelling_exemptor` is lacking, both in the codebase and on the GitHub wiki. Given the complexity of `yelling exemptor` (in the sense that it is a decorator with also a wrapping function), this is a bit of a problem. I might try to write something up this weekend if no-one beats me to it.

Once again, thank you for having a look into this. If you want to have a go at making certain commands yell, I'd suggest looking `haiku.py`, specifically how it responds to haikus:
```python
        if message.channel == discord.utils.get(
            self.bot.uqcs_server.channels, name=YELLING_CHANNEL_NAME
        ):
            await message.reply(f"Nice haiku:\n{haiku}".upper())
        else:
            await message.reply(f"Nice haiku:\n{haiku}")
```
Hopefully it should just be about going through all the interesting commands and doing something similar. Feel free to assign yourself to this issue if you want to have a crack.

andrewj-brown:
Ah, not quite (yet another example of how the `@yelling_exemptor` is confusing and silly).

What it actually does is checks through every argument on the command it's attached to, and then if its a string, requires it to be UPPERCASE in the yelling channel. Effectively, ensures that someone can't `/echo lower case text` - they have to `/echo UPPER CASE TEXT` or else uqcsbot will still :disapproval: them.

PR #164, opened 2023-10-02
Simplify regex in yelling.py

bradleysigma:
Noticed the superfluous regex in https://www.youtube.com/live/6wFATYjMTQE?t=1376

### DIFF
--- a/uqcsbot/yelling.py
+++ b/uqcsbot/yelling.py
@@ -166,7 +166,7 @@ class Yelling(commands.Cog):
         )
 
         # slightly more permissive version of discord's url regex, matches absolutely anything between http(s):// and whitespace
-        for url in re.findall(r"https?:\/\/[^\s]+", text, flags=re.UNICODE):
+        for url in re.findall(r"https?:\/\/\S+", text, flags=re.UNICODE):
             try:
                 resp = urlopen(url)
             except (ValueError, URLError):


andrewj-brown:
This is honestly a little disappointing. You're a regular contributor. We have 18 open issues you could have put work into, and you yourself have an open PR needing changes (granted, unlikely to be merged anyway, for committee reasons, but still). Why are you wasting my review time and your coding time with a 3-character change? You're perfectly capable of, y'know, fulfilling the actual purpose of hacktoberfest and making positive contributions to the bot.

Hell, issue #160 is right up your alley, considering how much work you've put into the yelling cog. Why wouldn't you go do that?

If you want a concrete reason why I'm not merging this...
- it's polluting the git blame, and making it unclear who's fault this regex is (not that it's a particularly complicated regex, but on principle, git blame should always be useful & accurate)
- it's using highly python-specific regex syntax, which actually makes it *less* easily readable than using the language-agnostic [^\s] construct - we're not code golfing here
- it's sending me stupid emails and interrupting me on a day when I need to be doing actual work on my assignments
- and, it's sending a poor message to external or future contributors that we're rubber-stamping hacktoberfest bullshit PRs from people in the "in group", making it harder to argue the case that we shouldn't accept them next time someone else submits one.

PR #166, opened 2023-10-07
Create Cato funciton 

bradleysigma:
As is tradition.
Some members of UQCS have started and finished university since the last time a Technology One _Big Game_ was held.
There is but one issue in the way of UQCS winning Big Game in 2024.
The purpose of this change is to remind @jenseni-git of his presidential duties.


JamesDearlove:
This joke is now over 4 years old, and I'd wager that the majority of current student members would have no clue what this is talking about.

Not merging it

PR #168, opened 2023-10-09
Remove Ununimplemented

bradleysigma:
Went to reimplement something from the Slackbot, only to find about the third of the so called unimplemented commands were no longer so.
- `calendar.py` - Largely implemented by `/whatsdue`;
- `channel_log.py` - Unneeded as channels creation is less dynamic on the Discord compared to the Slack;
- `coin.py` - Implemented by by `/coin`;
- `dominos.py` - Implemented by `/dominoscoupons`;
- `emoji_log.py` - Unneeded as emoji creation is less dynamic on the Discord compared to the Slack;
- `emojify.py` - Impractical to implement on the Discord due to the restricted emoji slots;
- `history.py` - Pinning culture is different on the Discord than on the Slack, and so won't need transferring. Also, this never really worked any way. It could be reimplemented into the Starboard, I guess;
- `id.py` - Unneeded as getting a use ID is much simpler on Discord;
- `link.py` - Brute force interface to the bot's database, unneeded;
- `pastexams.py` - Implemented by `/pastexams`;
- `wavie.py` - Discord prompts users waving to new members;
- `whoami.py` - Similar to `id.py`, unneeded;
- `xkcd.py` - Implemented by `\xkcd`.


bradleysigma:
Also removed `cookbook.py`, as it's implemented by `/repo find`.

JamesDearlove:
I'm actually of the opinion that we should remove the entire unimplemented folder, as was previously discussed internally. 

Now the bot is pretty much running entirely on the newer application command system, creating new commands looks a bit different to the Slack bot and any remaining useful commands might benefit from a rewrite to support Discord's unique features. Any commands that might be useful can be migrated to issues and linked to this repos history or directly to the Slack bot history

bradleysigma:
FWIW, after the Slackbot migrated from JavaScript to Python, there were a whole bunch of commands that weren't initially transferred, and similarly chucked into an unimplemented folder. An issue was made for these commands as a whole, but after a year or so, the last eleven commands had their own individual issues created.

bradleysigma:
Also, the remaining non-Discord/Slack Python code is of variable usefulness; the URL that `ascii.py` uses, for example, no longer exists.

andrewj-brown (approved):
LGTM. As discussed, unimplemented has little value anymore, and Jimmy has an excellent point in that everything still exists at https://github.com/UQComputingSociety/uqcsbot-slack/ anyway.

PR #169, opened 2023-10-15
Added logging for holidays cog

JamesDearlove:
Attempting to work out what is wrong with the holidays cog randomly failing. This adds logging for when the cog is run, so we can trace them on prod


andrewj-brown (approved):
ok jimothy

PR #170, opened 2023-10-23
Create ROT13 Command

bradleysigma:
Create [ROT 13](https://en.wikipedia.org/wiki/ROT13) command. Semi-imported from the Slack `!caesar` command. Can be used as an alternative to `||spoilers||`, with the distinction that the general shape of the message can be seen e.g. below the first use is identifiable as enciphering a URL. Has a context menu option, which responds with an ephemeral message; only shows the (probable) decryption to the user that called it, instead of publishing to the entire channel. Discord bots are limited to five context menu listings (for some reason), so either the context menu version of Zalgo or Mock (or both) would have to be removed for this to be added; both are currently commented out, but could uncomment one if there is a consensus. This has a meaningful usecase for the context menu version (to privately decrypt a version encrypted with the slash version), which Zalgo is kinda lacking.

image


bradleysigma:
Another difference it has to `||spoiler||` is that for this, you have the option to use an external tool to decode the message word by word, if you wish to attempt to see where a message is going and possibly cut yourself off part way though, instead of only having the option to reveal the entire message as a whole.

uqcsbot/text.py
```diff
@@ -33,17 +33,30 @@ async def encoding_autocomplete(
...
-        self.zalgo_menu = app_commands.ContextMenu(
-            name="Zalgo",
-            callback=self.zalgo_context,
-        )
-        self.bot.tree.add_command(self.zalgo_menu)
 
-        self.mock_menu = app_commands.ContextMenu(
-            name="Mock",
-            callback=self.mock_context,
+        ##        self.zalgo_menu = app_commands.ContextMenu(
```
49Indium, line 37: I'm impartial, as I don't use the context menu much. Potentially, the chosen options should be decided or at least approved by committee.

uqcsbot/text.py
```diff
@@ -273,6 +286,44 @@ async def zalgo_command(self, interaction: discord.Interaction, text: str):
 
         await interaction.response.send_message(self.zalgo_common(text))
 
+    def rot_13_cipher(self, text: str) -> str:
```
49Indium, line 289: I'm a bit surprised that modulo arithmetic wasn't used, but this implementation is quite readable.

uqcsbot/text.py
```diff
@@ -273,6 +286,44 @@ async def zalgo_command(self, interaction: discord.Interaction, text: str):
...
+    @app_commands.command(name="rot_13")
+    @app_commands.describe(text="Input text")
+    @yelling_exemptor()
+    async def rot_13_command(self, interaction: discord.Interaction, text: str):
+        """
+        Encodes the given text with the cunning ROT13 Cipher
+        """
+        await interaction.response.send_message(self.rot_13_cipher(text))
+
+    ##    async def rot_13_context(
```
49Indium, line 309: I think deciding between these two options really comes down to which gets used more often. As we don't have much use of ROT13 already, I can't really tell. I would think that having it ephemeral would be preferable, as people already have the command if they want everyone  to see it.

bradleysigma: The non-ephemeral version was written before I found out about the limit of five context commands.

uqcsbot/text.py
bradleysigma: Modular arithmetic means having to normalise to "a"=0, then do the modulo, then denormalise, which to me seems messier than this implementation
```python
def rot_13_cipher(self, text: str) -> str:
    result = ""
    for c in text:
        if "a" <= c <= "z":
            result += chr((ord(c) - ord("a") + 13) % 26 + ord("a"))
        elif "A" <= c <= "Z":
            result += chr((ord(c) - ord("A") + 13) % 26 + ord("A"))
        else:
            result += c
    return result
```
IIRC, I used modular arithmetic for the original `!caesar` command, as that had a variable rotation.

andrewj-brown (approved):
LGTM, pending discussion on context menu.

I'll add it to our infra to-do-list; we're currently *at* the limit of 5 commands (3 from text.py, plus the mod-only starboard blacklist/whitelist commands).

bradleysigma:
> I'll add it to our infra to-do-list

What's longer, the Infra to-do list, or @LimaoC's Outgoing Secretary to-do list?

LimaoC:
> @LimaoC's Outgoing Secretary to-do list

No such list exists, this is a fabricated lie made up by @bradleysigma

bradleysigma:
But here you give confirmation of the list's existence; were you lying then or are you lying now?
image

andrewj-brown:
> > I'll add it to our infra to-do-list
> 
> What's longer, the Infra to-do list, or @LimaoC's Outgoing Secretary to-do list?

Presumably the infra list, which is currently at 23 items (plus sub-items).

Our plan is to hold the meeting sometime within a month, so hopefully this PR doesn't have to wait too long.

PR #176, opened 2023-11-08
fix permission checks on starboard interactions & manage_cogs

andrewj-brown:
The default_permissions() decorator lets DMs through. This is cringe - we want to use default_permissions so that the server's role & moderation processes are abstracted out of the bot's code, and future committees can just fiddle with discord settings.

A more long-term solution is to have a decorator, `@requires_server` or the like, so we can continue to use default_permissions and just require that some commands must be used in the server. (This also would fix several other DM-based errors this PR identifies.) However, for now, let's just switch them all to `checks.has_permissions()` instead.

`grep -r permission` shows the following cogs use permissions:
- member_count: errors out on DMs anyway (should be fixed but who cares)
- minecraft: uses has_permissions() and Member-type-checks anyway
- remindme: has manual permission checks, these will throw `AttributeError` in DMs (should be fixed but who cares)
- starboard: fixed
- manage_cogs: fixed

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -851,13 +851,13 @@ files = [
 
 [[package]]
 name = "pyright"
-version = "1.1.316"
+version = "1.1.334"
 description = "Command line wrapper for pyright"
 optional = false
 python-versions = ">=3.7"
 files = [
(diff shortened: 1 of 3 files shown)


andrewj-brown:
Using mega gamer admin powers to merge instantly. it's a 3-line change, and I don't want people `/manage_cogs unload`'ing shit while this PR waits.

...also fuck typechecking

PR #185, opened 2023-11-30
Fixed the bug crashing leaderboards from anonymous users

49Indium:
The name provided by the AOC JSON is null if the user is anonymous. This is not corrected (and pyright doesn't pick up this error, as JSON is typed as `Dict[str, Any]`), so causes an error when displaying leaderboards of an anonymous user. this just make people have the name "Anon" if they don't have an AOC name.


JamesDearlove (approved):
lgtm

PR #187, opened 2023-12-03
Add allowed_mentions to ping messages

JamesDearlove:
As we disable the `allowed_mentions` on bot startup, messages that could ping people need to be explicitly defined. This patch adds these in to the advent cog.


andrewj-brown (approved):
LGTM!

I assume there's no way to make the roles more granular than "yes/no" - if there is, we should do that.

PR #188, opened 2023-12-14
Unwhitelist command

fattyhope:
New command for the new UQCS Minecraft Server. The command will "unwhitelist" the user while simultaneously kicking them from the server.

### DIFF
--- a/uqcsbot/minecraft.py
+++ b/uqcsbot/minecraft.py
@@ -98,9 +98,6 @@ class Minecraft(commands.Cog):
     async def mcunwhitelist(self, interaction: discord.Interaction, username: str):
         """Removes a username from the whitelist for the UQCS server."""
         db_session = self.bot.create_db_session()
-        query = db_session.query(MCWhitelist).filter(
-            MCWhitelist.discord_id == interaction.user.id
-        )
         is_user_admin = (
             isinstance(interaction.user, Member)
             and interaction.user.guild_permissions.manage_guild
@@ -126,11 +123,7 @@ class Minecraft(commands.Cog):
 
             # If the responses indicate successful removal, remove from the database item
             if "Removed" in response_remove[0]:
-                query = (
-                    db_session.query(MCWhitelist)
-                    .filter(MCWhitelist.mc_username == username)
-                    .delete()
-                )
+                db_session.query(MCWhitelist).filter(MCWhitelist.mc_username == username).delete()
                 db_session.commit()
 
                 await self.bot.admin_alert(
--- a/uqcsbot/minecraft.py
+++ b/uqcsbot/minecraft.py
@@ -123,7 +123,9 @@ class Minecraft(commands.Cog):
 
             # If the responses indicate successful removal, remove from the database item
             if "Removed" in response_remove[0]:
-                db_session.query(MCWhitelist).filter(MCWhitelist.mc_username == username).delete()
+                db_session.query(MCWhitelist).filter(
+                    MCWhitelist.mc_username == username
+                ).delete()
                 db_session.commit()
 
                 await self.bot.admin_alert(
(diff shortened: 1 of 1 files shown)


uqcsbot/minecraft.py
```diff
@@ -93,6 +93,54 @@ async def mcwhitelist(self, interaction: discord.Interaction, username: str):
 
         db_session.close()
 
+    @app_commands.command()
+    @app_commands.describe(username="Minecraft username to unwhitelist.")
+    @yelling_exemptor(input_args=["username"])
```
andrewj-brown, line 98: I wouldn't bother adding the `@yelling_exemptor`, that just makes the command unusable in #yelling (it'll enforce that the username needs to be all-upper-case, which probably mucks with the actual whitelisting)

uqcsbot/minecraft.py
```diff
@@ -93,6 +93,54 @@ async def mcwhitelist(self, interaction: discord.Interaction, username: str):
...
+        query = db_session.query(MCWhitelist).filter(
+            MCWhitelist.discord_id == interaction.user.id
+        )
+        is_user_admin = (
+            isinstance(interaction.user, Member)
+            and interaction.user.guild_permissions.manage_guild
+        )
+
+        # If the user has already whitelisted someone, and they aren't an admin deny it.
+        if not is_user_admin and query.count() > 0:
```
andrewj-brown, line 111: This check is dodgy - it should just be `not is_user_admin`. Otherwise someone who's whitelisted an account can ban anyone they want

andrewj-brown: Subsequently, I guess you could just put `@app_commands.checks.has_permissions(manage_guild=True)` on the whole command and remove the explicit permission check, but that's less of a concern.

uqcsbot/minecraft.py
```diff
@@ -93,6 +93,54 @@ async def mcwhitelist(self, interaction: discord.Interaction, username: str):
...
+                db_session.commit()
+
+                await self.bot.admin_alert(
+                    title="Minecraft Server Unwhitelist",
+                    description=f"{response_remove[0]}\n{response_kick[0]}",
+                    footer=f"Action performed by {interaction.user}",
+                    colour=Colour.red(),  # Use red to indicate removal
+                )
+
+            await interaction.response.send_message(
```
andrewj-brown, line 138: This message could be beefed up a little bit, because it's realistically an error message - it couldn't kick or couldn't un-whitelist the player for some reason.

uqcsbot/minecraft.py
```diff
@@ -119,7 +167,6 @@ async def mcadmin(self, interaction: discord.Interaction, command: str):
             for split in split_response[1:]:
                 await interaction.followup.send(f"```{split}```")
 
-        # Just to be safe, send this to the admin log channel as well.
```
andrewj-brown, line 122: If you could re-add this comment (I assume removing it was just a mistake) that'd be nice.

andrewj-brown (changes requested):
Looks good - I had no idea what you were working on, but this is a really useful PR. Got a couple changes, but they're all pretty minor; this will easily be mergeable tonight.

uqcsbot/minecraft.py
fattyhope: Confused about this, lines 135 to 138 will just send a message to notify an individual has been kicked and unwhitelisted. 

error

andrewj-brown: Hm, right. It was outside the check on line 127, which made me think it was an error case, but on second look it is just the normal message.

In that case, consider this a request for the bot to say/do something specific *if* the line 127 check fails, so we know if the bot fails to kick/fails to unwhitelist.

fattyhope: After a bit of testing, I've just removed it and get it to display the remove_response[0] since removing a non-existence user will display results shown below:
image

uqcsbot/minecraft.py
fattyhope: Added this back in, thanks for catching that

andrewj-brown (approved):
LGTM and merging. Couple extra things this PR has made me think of, but they're worth doing separately

PR #190, opened 2023-12-20
Added command to get current MC server player count

Antimatter543:
*Do /mcplayers to get player count, and current users on the minecraft server

Currently hard-coded the ip into it, this is cursed but I can't tell if RCON IP is the same as public IP.


uqcsbot/minecraft.py
```diff
@@ -48,6 +49,27 @@ async def send_rcon_command(self, command: str):
 
         return response
 
+    @app_commands.command()
+    async def mcplayers(self, interaction: discord.Interaction):
+        """Returns the number and list of people currently playing on the Minecraft server."""
+        server = JavaServer.lookup(
+            "minecraft.uqcs.org:25605"
```
andrewj-brown, line 56: Alright I have confirmed this is different from the RCON addr. Probably don't hard-code it - add a new environment variable. You should be able to copy how the other MC commands do it.

uqcsbot/minecraft.py
```diff
@@ -48,6 +49,27 @@ async def send_rcon_command(self, command: str):
...
+        )  # Does this need to be hard coded?? Is RCON addr/IP the same?
+        status = server.status()  # type: ignore
```
andrewj-brown, line 58: Note for posterity. I'm conditionally approving this type:ignore, because it's part of a library. Will be following up and checking why, but that doesn't have to block this PR - I think it's to do with `mcserver` using Any type instead of actual object types.

andrewj-brown (changes requested):
Review as promised. If you fix the hard-coded IP and switch it out for an env variable, I reckon it's all good.

andrewj-brown (approved):
LGTM. You'll need to merge the upstream changes into your branch before I merge - there should be a button down the bottom of the PR that says "update branch", and it should be able to autoresolve.

Issue #165, opened 2023-10-02
uqcsbot poll


jenseni-git:
New cog idea: poll.

/poll will take in between 3 and 11 arguments.
- The first argument will be the polling question to be proposed to voters
- The second and third (as a minimum) up to the 11th argument will be the polling answers for users to select.
- When the function is called, the bot will send the argument question, then format each possible response with the relevant numerical emoji 1-n (0 being 10) displayed next to the relevant answer option. The bot will then react with the relevant numerical emojis.

Possible pain points:
- Message overflow (Should this be handled by multiple messages or an error; how may this effect the below idea).

Extensibility idea: an addtional 'ping me' argument of a time. At the time as inputted, the bot will ping the sender again, and reference the original poll to remind the poller that the poll is now finished.

andrewj-brown:
Potentially a good use-case for using interaction buttons instead of reactions, which is the way that bots are "supposed" to do stuff like this.

49Indium:
Closing this, as [Discord is rolling out a native poll feature](https://support.discord.com/hc/en-us/articles/22163184112407-Polls-FAQ).

PR #193, opened 2024-03-04
Add poll command

49Indium:
Adds a poll command that allows for multi-option polls via reacts as mentioned in #165.

Originally, I was going to use buttons as @andrewj-brown suggested, but as these are displayed identically to everyone, without ephemeral madness, I could not show which choice you had selected for the poll. Better to keep it simple and use reacts, which are pretty ideal for this case (e.g. you can see who voted for what, vote total, etc.).

I did end up using a modal for creating the poll options, as this is a bit neater than entering into 11 different arguments.

I'm going to leave the ping to `/remindme`, but happy to add it in if it's still wanted.

### DIFF
--- a/uqcsbot/poll.py
+++ b/uqcsbot/poll.py
@@ -1,0 +1,82 @@
+import discord
+from discord import app_commands
+from discord.ext import commands
+
+from typing import Any
+
+# The unicode points for each number emoji starting from 0 to 10
+EMOJI_NUMBERS = [
+    "\u0030\u20E3",
+    "\u0031\u20E3",
+    "\u0032\u20E3",
+    "\u0033\u20E3",
+    "\u0034\u20E3",
+    "\u0035\u20E3",
+    "\u0036\u20E3",
+    "\u0037\u20E3",
+    "\u0038\u20E3",
+    "\u0039\u20E3",
+    "\U0001F51F",
+]
+
+
+class PollMenu(discord.ui.Modal, title="New Poll"):
+    """The menu to create a new poll"""
+
+    poll_question: discord.ui.TextInput[Any] = discord.ui.TextInput(
+        label="A question of the poll",
+        placeholder="Your question goes here (e.g. What does UQCS stand for?)",
+        max_length=100,
+    )
+    poll_options: discord.ui.TextInput[Any] = discord.ui.TextInput(
+        label="The options for the poll (one per line)",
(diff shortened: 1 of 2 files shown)


uqcsbot/poll.py
```diff
@@ -0,0 +1,82 @@
...
+    "\u0032\u20E3",
+    "\u0033\u20E3",
+    "\u0034\u20E3",
+    "\u0035\u20E3",
+    "\u0036\u20E3",
+    "\u0037\u20E3",
+    "\u0038\u20E3",
+    "\u0039\u20E3",
+    "\U0001F51F",
+]
```
JamesDearlove, line 20: Imo, rather than unicode these could just be the the Discord emoji names `one`, `two`, `three`, etc.

uqcsbot/poll.py
```diff
@@ -0,0 +1,82 @@
...
+class PollMenu(discord.ui.Modal, title="New Poll"):
+    """The menu to create a new poll"""
+
+    poll_question: discord.ui.TextInput[Any] = discord.ui.TextInput(
+        label="A question of the poll",
+        placeholder="Your question goes here (e.g. What does UQCS stand for?)",
+        max_length=100,
+    )
+    poll_options: discord.ui.TextInput[Any] = discord.ui.TextInput(
+        label="The options for the poll (one per line)",
```
JamesDearlove, line 32: Might be good to include the max 10 in the dialog, as you don't find out til after you put more than 10

uqcsbot/poll.py
```diff
@@ -0,0 +1,82 @@
...
+        style=discord.TextStyle.long,
+        placeholder="Option 1 (e.g. UQ Computing Society)\nOption 2 (e.g. UQ Caveman Society)\nOption 3...",
```
JamesDearlove, line 34: Comment: I believe we actually refer to it as the UQ Cavepainting Society

JamesDearlove (commented):
I was playing with this, got a couple of comments. Was also wondering if this would look better styled as an embed message. Here's an example: 
image

JamesDearlove:
Also could look better with a standard message and Discord's now built in markdown

49Indium:
Given that [Discord is rolling out a polls feature](https://support.discord.com/hc/en-us/articles/22163184112407-Polls-FAQ), I'm going to close this.

PR #194, opened 2024-03-06
Added typing to command_utils.py

49Indium:
Added types to `utils/command_utils.py` to slowly help complete #134. I did have to use `Any` types twice for the decorator. Once is for `*args`, which seems reasonable. However, I had to also use it for `self`, which seems to be the easiest solution, given https://stackoverflow.com/a/69528375. I'd like to hear what others think.

I also removed the check for the bot, as the type system recommends that this is no longer necessary.


andrewj-brown:
complete direction-change: command_utils.py is no longer necessary, as all of the bot's commands use interactions instead (which can be deferred with a "thinking" message as part of Discord's UI, instead of using reactions). The only place it's used is the unimplemented/ folder.

Could you just delete the file instead?

49Indium:
I had thought about this when I was editing it... should have thought harder

andrewj-brown (commented):
ezpz

andrewj-brown (approved):
mfw i forget to hit "approve"

PR #196, opened 2024-03-09
Add lichen cat

Jake-Moss:


49Indium (commented):
I don't know the joke, but I think that this is a neat idea! I'll let someone else decide how/if this fits in with UQCS lore. Potentially not mentioning lichen cat within the command description would be funnier (whilst still noting it in the code).

Additionally, I prefer the style that Moss cat uses with `"\n".join()` over having `+ "\n"` after each new line, but I'm not too fussed about this either way (the raw strings are great though).

Jake-Moss:
> I don't know the joke, but I think that this is a neat idea! I'll let someone else decide how/if this fits in with UQCS lore. Potentially not mentioning lichen cat within the command description would be funnier (whilst still noting it in the code).

I believe lichen is the OpenBSD server for Operating Systems Architecture (COMP3301)

> Additionally, I prefer the style that Moss cat uses with `"\n".join()` over having `+ "\n"` after each new line, but I'm not too fussed about this either way (the raw strings are great though).

Done, I also forgot the code fences so I've added them

andrewj-brown (approved):
*technically*, OS'es doesn't use Lichen - they use BSD, but on a range of VMs that students have root access to. I'm unsure if anything actually uses Lichen itself; more commonly it's used as an alternative proxy server when Moss gets slow.

Also LGTM.

Issue #198, opened 2024-03-19
Potentially inappropriate #yelling message


Ninjaman10p:
In #yelling, UQCSBot chides people who forget (or fail) to yell. However, one of its messages (attached)
![image](https://github.com/UQComputingSociety/uqcsbot-discord/assets/59110523/abc24be9-239e-4bdc-bffb-7634bee874ed)
has always seemed to me like it might be somewhat inappropriate for a club-sanctioned bot to be saying.

I would argue that this message risks making some people feel unwelcome, and seems out of place with the other messages (which are mostly light-hearted mocking or puns), given that this one directly (even if ironically) calls for violence.

PR #199, opened 2024-03-19
Removed potentially inappropriate #yelling response

Ninjaman10p:
As I mentioned in https://github.com/UQComputingSociety/uqcsbot-discord/issues/198, one of the messages in the #yelling responses includes an (ironic) call for violence which may make some members uncomfortable and seems inappropriate for a club-sanctioned bot to be saying.

This pull request removes that message.


JamesDearlove (approved):
I agree with the removal of this yelling response.

Thanks for the PR @Ninjaman10p . I invite you to add a new response to fill its place, but not mandatory if you'd prefer not to.

PR #202, opened 2024-06-15
Remove the unimplemented folder

JamesDearlove:
This was discussed a while ago but was never actioned on. This is to remove the unimplemented folder, as most things in here are either no longer viable or too different from the Slack. 

You might ask, why remove it? Well we already have the Slack bot repo for inspiration, so we really don't need the duplicate here. If desired, we could add some of these as issues as todo items but let new contributors create the implementation from scratch


49Indium (approved):
Looks good to me

PR #203, opened 2024-06-16
Adding CODEOWNERS File

49Indium:
To help contributors find people knowledgeable with each part of the bot and to try to simplify the getting reviews, I've added a [CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#about-code-owners) file. I used a combination of `git shortlog -n`, `git blame` and my own knowledge to assign code owners.

I've checked with everyone on this file, but feel free to add your name in spots if you think it's missing. There are a few files without codeowners, but I think that this is ok. These are usually because the main contributor didn't want to be the codeowner, or because the file is very old.

I'll also add documentation for this as well, but that is for the github pages, so will be a different PR.


jenseni-git:
I'm assuming the current errors are because people were kicked after 2FA?

jenseni-git (changes requested):
Consider changing offending tags from @XXXX to @ XXXX. Otherwise, I think it's valuable documentation.

Issue #204, opened 2024-06-30
/dominoscoupons Should Use FrugalFeeds


49Indium:
[As brought up by lucid8426 on discord](https://discord.com/channels/813324385179271168/836243768411160606/1256604024501964955), the `/dominoscoupons` command provides discounts that may not work. Potentially we should also include [FrigralFeeds](https://www.frugalfeeds.com.au/) coupons.

One way to do this would be to add an optional argument for the source. I'm unsure what should happen by default: either just selecting one of the sources or trying to combine them. Ideally we would have some way to test if the coupons work, but I'm unsure if that is achievable and isn't required.

Note that we should probably still keep `https://www.couponese.com/store/dominos.com.au/` in some form, as a good back-up as some of the coupons still work.

tj-dodd:
I've made decent progress on this issue. The user is now able to specify whether they want a list of codes from FrugalFeeds or Couponese through an optional argument, and by default the user is provided with a combined list. Right now the combined list is a random selection of discount codes (as you end up with a pretty big list), and I'm thinking of implementing a button the user can press to refresh the combined list - although what I've got might work well enough as is.

49Indium:
Completed with #205

PR #205, opened 2024-06-30
Choose between FrugalFeeds and Couponese (issue #204)

tj-dodd:
This gives the user the ability to choose between FrugalFeeds and Couponese through a source optional argument. I ended up making it so the bot's default behaviour is to give the user a randomly generated combined list of coupons. I changed the hyperlink for the embed to FrugalFeeds, as there's a more comprehensive collection of coupon codes there.

### DIFF
--- a/uqcsbot/dominos_coupons.py
+++ b/uqcsbot/dominos_coupons.py
@@ -1,9 +1,10 @@
 from datetime import datetime
-from typing import List
+from typing import List, Dict
 import logging
 import requests
 from requests.exceptions import RequestException
 from bs4 import BeautifulSoup
+import random
 
 import discord
 from discord import app_commands
@@ -12,9 +13,9 @@ from discord.ext import commands
 from uqcsbot.bot import UQCSBot
 from uqcsbot.yelling import yelling_exemptor
 
-
 MAX_COUPONS = 10  # Prevents abuse
 COUPONESE_DOMINOS_URL = "https://www.couponese.com/store/dominos.com.au/"
+FRUGAL_FEEDS_DOMINOS_URL = "https://www.frugalfeeds.com.au/dominos/"
 
 
 class HTTPResponseException(Exception):
@@ -22,9 +23,10 @@ class HTTPResponseException(Exception):
     An exception for when a HTTP response is not requests.codes.ok
     """
 
-    def __init__(self, http_code: int, *args: object) -> None:
+    def __init__(self, http_code: int, url: str, *args: object) -> None:
         super().__init__(*args)
         self.http_code = http_code
+        self.url = url
 
 
 class DominosCoupons(commands.Cog):
@@ -36,6 +38,7 @@ class DominosCoupons(commands.Cog):
         number_of_coupons="The number of coupons to return. Defaults to 5 with max 10.",
         ignore_expiry="Indicates to include coupons that have expired. Defaults to True.",
         keywords="Words to search for within the coupon. All coupons descriptions will mention at least one keyword.",
+        source="Website to source coupons from (couponese or frugalfeeds). Defaults to both.",
     )
     @yelling_exemptor(input_args=["keywords"])
     async def dominoscoupons(
@@ -44,6 +47,7 @@ class DominosCoupons(commands.Cog):
         number_of_coupons: int = 5,
         ignore_expiry: bool = True,
         keywords: str = "",
+        source: str = "",
     ):
         """
         Returns a list of dominos coupons
@@ -54,19 +58,31 @@ class DominosCoupons(commands.Cog):
                 ephemeral=True,
             )
             return
+
+        if source:
+            if source != "couponese" and source != "frugalfeeds":
+                await interaction.response.send_message(
+                    content=f"That website isn't recognised. Try couponese or frugalfeeds.",
+                    ephemeral=True,
+                )
+                return
+
         await interaction.response.defer(thinking=True)
 
         try:
-            coupons = _get_coupons(number_of_coupons, ignore_expiry, keywords.split())
+            coupons = _get_coupons(
+                number_of_coupons, ignore_expiry, keywords.split(), source
+            )
         except RequestException as error:
             resp_content = (
                 error.response.content if error.response else "No response error given."
             )
+            request_url = error.request.url if error.request else "Unknown site."
             logging.warning(
-                f"Could not connect to dominos coupon site ({COUPONESE_DOMINOS_URL}): {resp_content}"
+                f"Could not connect to dominos coupon site ({request_url}): {resp_content}"
             )
             await interaction.edit_original_response(
-                content=f"Sadly could not reach the coupon website (<{COUPONESE_DOMINOS_URL}>)..."
+                content=f"Sadly could not reach the coupon website (<{request_url}>)..."
             )
             return
         except HTTPResponseException as error:
@@ -74,19 +90,19 @@ class DominosCoupons(commands.Cog):
                 f"Received a HTTP response code {error.http_code}. Error information: {error}"
             )
             await interaction.edit_original_response(
-                content=f"Could not find the coupons on the coupon website (<{COUPONESE_DOMINOS_URL}>)..."
+                content=f"Could not find the coupons on the coupon website {error.url}..."
             )
             return
 
         if not coupons:
             await interaction.edit_original_response(
-                content=f"Could not find any coupons matching the given arguments from the coupon website (<{COUPONESE_DOMINOS_URL}>)."
+                content=f"Could not find any coupons matching the given arguments from the coupon website/s."
             )
             return
 
         embed = discord.Embed(
             title="Domino's Coupons",
-            url=COUPONESE_DOMINOS_URL,
+            url=FRUGAL_FEEDS_DOMINOS_URL,
             description=f"Keywords: *{keywords}*" if keywords else None,
             timestamp=datetime.now(),
         )
@@ -123,12 +139,14 @@ class Coupon:
         return keyword.lower() in self.description.lower()
 
 
-def _get_coupons(n: int, ignore_expiry: bool, keywords: List[str]) -> List[Coupon]:
+def _get_coupons(
+    n: int, ignore_expiry: bool, keywords: List[str], source: str
+) -> List[Coupon]:
     """
     Returns a list of n Coupons
     """
 
-    coupons = _get_coupons_from_page()
+    coupons = _get_coupons_from_page(source)
 
     if not ignore_expiry:
         coupons = [coupon for coupon in coupons if coupon.is_valid()]
@@ -139,27 +157,88 @@ def _get_coupons(n: int, ignore_expiry: bool, keywords: List[str]) -> List[Coupo
             for coupon in coupons
             if any(coupon.keyword_matches(keyword) for keyword in keywords)
         ]
+
+    if not source:
+        random.shuffle(coupons)
+        return coupons[:n]
+
     return coupons[:n]
 
 
-def _get_coupons_from_page() -> List[Coupon]:
-    """
-    Strips results from html page and returns a list of Coupon(s)
-    """
-    http_response = requests.get(COUPONESE_DOMINOS_URL)
-    if http_response.status_code != requests.codes.ok:
-        raise HTTPResponseException(http_response.status_code)
-    soup = BeautifulSoup(http_response.content, "html.parser")
-    soup_coupons = soup.find_all(class_="ov-coupon")
-
+def _get_coupons_from_page(source: str) -> List[Coupon]:
+    urls: List[str] = []
     coupons: List[Coupon] = []
 
-    for soup_coupon in soup_coupons:
-        expiry_date_str = soup_coupon.find(class_="ov-expiry").get_text(strip=True)
-        description = soup_coupon.find(class_="ov-desc").get_text(strip=True)
-        code = soup_coupon.find(class_="ov-code").get_text(strip=True)
-        coupon = Coupon(code, expiry_date_str, description)
-        coupons.append(coupon)
+    website_coupon_classes: Dict[str, Dict[str, str]] = {
+        COUPONESE_DOMINOS_URL: {
+            "expiry": "ov-expiry",
+            "description": "ov-desc",
+            "code": "ov-code",
+        },
+        FRUGAL_FEEDS_DOMINOS_URL: {
+            "expiry": "column-3",
+            "description": "column-2",
+            "code": "column-1",
+        },
+    }
+
+    if source == "couponese":
+        urls.append(COUPONESE_DOMINOS_URL)
+    elif source == "frugalfeeds":
+        urls.append(FRUGAL_FEEDS_DOMINOS_URL)
+    else:
+        urls = [FRUGAL_FEEDS_DOMINOS_URL, COUPONESE_DOMINOS_URL]
+
+    for url in urls:
+        http_response: requests.Response = requests.get(url)
+        if http_response.status_code != requests.codes.ok:
+            raise HTTPResponseException(http_response.status_code, url)
+
+        soup = BeautifulSoup(http_response.content, "html.parser")
+        soup_coupons: List[BeautifulSoup] = []
+
+        if url == COUPONESE_DOMINOS_URL:
+            soup_coupons = soup.find_all(class_="ov-coupon")
+        elif url == FRUGAL_FEEDS_DOMINOS_URL:
+            tables = soup.select('[class^="tablepress"]')
+            for table in tables:
+                # Headers have stuff we don't want
+                rows = table.find_all("tr")[1:]
+                soup_coupons.extend(rows)
+
+        siteclass: Dict[str, str] = website_coupon_classes.get(url, {})
+
+        for soup_coupon in soup_coupons:
+            expiry_date_container = soup_coupon.find(class_=siteclass.get("expiry"))
+            description_container = soup_coupon.find(
+                class_=siteclass.get("description")
+            )
+            code_container = soup_coupon.find(class_=siteclass.get("code"))
+
+            if (
+                not expiry_date_container
+                or not description_container
+                or not code_container
+            ):
+                continue
+
+            expiry_date_str: str = expiry_date_container.get_text(strip=True)
+            description: str = description_container.get_text(strip=True)
+            code: str = code_container.get_text(strip=True)
+
+            if url == FRUGAL_FEEDS_DOMINOS_URL:
+                date_values: List[str] = expiry_date_str.split()
+                try:
+                    # Convert shortened month to numerical value
+                    month: int = datetime.strptime(date_values[1], "%b").month
+                except (ValueError, IndexError):
+                    continue
+                expiry_date_str = "{year}-{month}-{day}".format(
+                    year=int(date_values[2]), month=month, day=int(date_values[0])
+                )
+
+            coupon = Coupon(code, expiry_date_str, description)
+            coupons.append(coupon)
 
     return coupons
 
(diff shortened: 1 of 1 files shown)


uqcsbot/dominos_coupons.py
```diff
@@ -139,27 +157,88 @@ def _get_coupons(n: int, ignore_expiry: bool, keywords: List[str]) -> List[Coupo
             for coupon in coupons
             if any(coupon.keyword_matches(keyword) for keyword in keywords)
         ]
-    return coupons[:n]
 
+    if not source:
```
49Indium, line 161: This is a great idea. Given the coupons are never ordered, the coupons should be shuffled regardless of `source`.

uqcsbot/dominos_coupons.py
```diff
@@ -139,27 +157,88 @@ def _get_coupons(n: int, ignore_expiry: bool, keywords: List[str]) -> List[Coupo
...
+
+        for soup_coupon in soup_coupons:
+            expiry_date_container = soup_coupon.find(class_=siteclass.get("expiry"))
+            description_container = soup_coupon.find(
+                class_=siteclass.get("description")
+            )
+            code_container = soup_coupon.find(class_=siteclass.get("code"))
+
+            if (
+                not expiry_date_container
```
49Indium, line 219: Potentially we might be able to still show coupons that don't have a date, but this is up to you.

uqcsbot/dominos_coupons.py
```diff
@@ -139,27 +157,88 @@ def _get_coupons(n: int, ignore_expiry: bool, keywords: List[str]) -> List[Coupo
...
+            description: str = description_container.get_text(strip=True)
+            code: str = code_container.get_text(strip=True)
+
+            if url == FRUGAL_FEEDS_DOMINOS_URL:
+                date_values: List[str] = expiry_date_str.split()
+                try:
+                    # Convert shortened month to numerical value
+                    month: int = datetime.strptime(date_values[1], "%b").month
+                except (ValueError, IndexError):
+                    continue
```
49Indium, line 235: Even if the coupon has an unidentifiable month, it should probably still be included (maybe with question marks for the month).

uqcsbot/dominos_coupons.py
```diff
@@ -139,27 +157,88 @@ def _get_coupons(n: int, ignore_expiry: bool, keywords: List[str]) -> List[Coupo
...
+                or not description_container
+                or not code_container
+            ):
+                continue
+
+            expiry_date_str: str = expiry_date_container.get_text(strip=True)
+            description: str = description_container.get_text(strip=True)
+            code: str = code_container.get_text(strip=True)
+
+            if url == FRUGAL_FEEDS_DOMINOS_URL:
```
49Indium, line 229: This one is up to you, but it might be nice to restructure the `Coupon` class to have the date as an `Optional[Datetime]` instead of a string (or maybe even a `Datetime|str`, where the Datetime is used by default, but if it can't be parsed, the `str` is used to still provide some information about the date). This would allow us to have all date parsing here, regardless of URL used.

49Indium (changes requested):
Firstly, thank you for contributing! This is a great improvement and should help reduce the trial and error needed when getting pizza!

Overall the command adjustments appear to to be working great. I like how you have structured `_get_coupons_from_page` and the idea of shuffling coupons each time. There are some small adjustments I think would be nice which I've attached with the code. Some of these comments are up to your discretion; they are all relatively minor and could be argued either way if they are needed.

uqcsbot/dominos_coupons.py
tj-dodd: I agree with this - I admit this was mostly a band-aid solution at the time as I was dealing with invalid date formats causing index errors.

uqcsbot/dominos_coupons.py
tj-dodd: I'll have a look into this, however the user is generally provided enough coupon codes where we can get away with excluding coupons that don't have a date. (and I'm not sure if I've come across a coupon code without a date yet)

uqcsbot/dominos_coupons.py
```diff
@@ -54,39 +58,51 @@ async def dominoscoupons(
...
+                number_of_coupons, ignore_expiry, keywords.split(), source
+            )
         except RequestException as error:
             resp_content = (
                 error.response.content if error.response else "No response error given."
             )
+            request_url = error.request.url if error.request else "Unknown site."
             logging.warning(
-                f"Could not connect to dominos coupon site ({COUPONESE_DOMINOS_URL}): {resp_content}"
+                f"Could not connect to dominos coupon site ({request_url}): {resp_content}"
```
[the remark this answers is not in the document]
tj-dodd: This is a great idea, I'll start on working on getting this implemented.

tj-dodd:
Just committed addressing your changes - if there's any issues let me know.

I ended up reverting back to how you were originally parsing dates. My additions were unnecessary I think - to be transparent, I'm not completely sure what I was trying to do, but I believe at one point I mistakenly assumed I had to do some extra parsing as the month formatting on FrugalFeeds is slightly different and I was encountering a couple of weird dates (e.g. soon!). I think leaving the date as a string if it's unable to be parsed is a much better implementation. There's really only 1-2 ambiguous dates in the FrugalFeeds list and the format is slightly different to Couponese, but the user still receives more than enough information even if the date is passed through as an unformatted string.

I'm still not certain if giving users a random sample of discount codes is the best way to approach things, as people might get unlucky (e.g. bot giving them 5 different mudcake discounts, excellent!), but I guess it's on the user to make use of the keywords function, and if you're getting decent results then I guess it's fine.

tj-dodd:
nevermind, I think most recent commit is the best way to do the formatting

tj-dodd:
Some small changes:

- Bot now hints user to change source if their keywords don't return any results on a single website
- FrugalFeeds went rogue and started advertising pizza hut deals on their dominos page, these have been filtered out
- Fixed FrugalFeeds formatting for discounts with multiple codes
- Moved hyperlinks down to description area, bot now links both FrugalFeeds and Couponese if both are selected

uqcsbot/dominos_coupons.py
```diff
@@ -44,50 +48,105 @@ async def dominoscoupons(
...
-            )
+            if source == "both":
+                content_str = "Could not find any coupons matching the given arguments from both websites."
+            else:
+                content_str = f"Could not find any coupons matching the given arguments from {source}. You can try changing the website through the `source` command."
+            await interaction.edit_original_response(content=content_str)
             return
 
+        if source == "both":
+            description_string = f"Sourced from [FrugalFeeds]({COUPONESE_DOMINOS_URL}) and [Couponese]({FRUGAL_FEEDS_DOMINOS_URL})"
```
49Indium, line 136: This has the links the wrong way around

uqcsbot/dominos_coupons.py
```diff
@@ -44,50 +48,105 @@ async def dominoscoupons(
...
         """
+
         if number_of_coupons < 1 or number_of_coupons > MAX_COUPONS:
             await interaction.response.send_message(
                 content=f"You can't have that many coupons. Try a number between 1 and {MAX_COUPONS}.",
                 ephemeral=True,
             )
             return
+
+        if source != "both":
```
49Indium, line 64: This shouldn't be needed any more, given that the source can only be "both", "couponese" or "frugalfeeds" (and should pass type checking).

uqcsbot/dominos_coupons.py
```diff
@@ -44,50 +48,105 @@ async def dominoscoupons(
...
-                f"Received a HTTP response code {error.http_code}. Error information: {error}"
-            )
-            await interaction.edit_original_response(
-                content=f"Could not find the coupons on the coupon website (<{COUPONESE_DOMINOS_URL}>)..."
-            )
-            return
+        coupons: List[Coupon] = []
+        stored_url: str = " "
+
+        for tries in range(0, NUMBER_WEBSITES):
```
49Indium, line 77: Instead of looping over a range, it might be better to loop over a list of the URLs you are going to gather coupons from. You appear to do this within `_get_coupons_from_page`,  so I'm unsure why multiple attempts are needed here.

uqcsbot/dominos_coupons.py
```diff
@@ -44,50 +48,105 @@ async def dominoscoupons(
...
-                content=f"Could not find any coupons matching the given arguments from the coupon website (<{COUPONESE_DOMINOS_URL}>)."
-            )
+            if source == "both":
+                content_str = "Could not find any coupons matching the given arguments from both websites."
+            else:
+                content_str = f"Could not find any coupons matching the given arguments from {source}. You can try changing the website through the `source` command."
+            await interaction.edit_original_response(content=content_str)
             return
 
+        if source == "both":
```
49Indium, line 135: I like the way that you have included this in the embed.

uqcsbot/dominos_coupons.py
```diff
@@ -123,12 +182,14 @@ def keyword_matches(self, keyword: str) -> bool:
...
-def _get_coupons(n: int, ignore_expiry: bool, keywords: List[str]) -> List[Coupon]:
+def _get_coupons(
+    n: int, ignore_expiry: bool, keywords: List[str], source: str
+) -> List[Coupon]:
     """
     Returns a list of n Coupons
     """
 
-    coupons = _get_coupons_from_page()
+    coupons: List[Coupon] = _get_coupons_from_page(source)
```
49Indium, line 192: The explicit type on this line isn't needed. Feel free to leave it if you think it makes it more readable.

uqcsbot/dominos_coupons.py
```diff
@@ -139,27 +200,95 @@ def _get_coupons(n: int, ignore_expiry: bool, keywords: List[str]) -> List[Coupo
...
             if any(coupon.keyword_matches(keyword) for keyword in keywords)
         ]
+
+    random.shuffle(coupons)
+
     return coupons[:n]
 
 
-def _get_coupons_from_page() -> List[Coupon]:
+def _get_coupons_from_page(source: str) -> List[Coupon]:
```
49Indium, line 209: It would be nice to use the literal type again for `source`. This should avoid some of your case checking below. Additionally, I'd personally make this function only accept a single source (so not "both"), and let `_get_coupons` turn multiple sources into multiple calls of this function.

uqcsbot/dominos_coupons.py
```diff
@@ -139,27 +200,95 @@ def _get_coupons(n: int, ignore_expiry: bool, keywords: List[str]) -> List[Coupo
...
+                or not code_container
+            ):
+                continue
+
+            expiry_date_str: str = expiry_date_container.get_text(strip=True)
+            description: str = description_container.get_text(strip=True)
+            code: str = code_container.get_text(strip=True)
+
+            temp_code: str = code.replace(",", "")
+            if not temp_code.isdigit():
```
49Indium, line 274: I'm assuming this is to remove the frugalfeed coupons that aren't for dominos.
However, this appears to remove some couponese coupons (the ones that have spaces in them) as well.
I think a comment here explaining why this occurs would be good.

uqcsbot/dominos_coupons.py
```diff
@@ -44,50 +48,105 @@ async def dominoscoupons(
         number_of_coupons: int = 5,
         ignore_expiry: bool = True,
```
49Indium, line 49: Unrelated to the changes you have made, but it might be best to make this `False` by default. This should provide a better list of coupons,  at the risk of having less coupons to print out.

uqcsbot/dominos_coupons.py
```diff
@@ -139,27 +200,95 @@ def _get_coupons(n: int, ignore_expiry: bool, keywords: List[str]) -> List[Coupo
...
+
+            expiry_date_str: str = expiry_date_container.get_text(strip=True)
+            description: str = description_container.get_text(strip=True)
+            code: str = code_container.get_text(strip=True)
+
+            temp_code: str = code.replace(",", "")
+            if not temp_code.isdigit():
+                continue
+
+            code = code.replace(",", " ")
```
49Indium, line 277: I think using commas to separate codes (rather than spaces) is better. This is just my personal opinion, so it's up to you.

49Indium (changes requested):
Thanks for making those changes. It's annoying that FrugalFeeds has included different codes.
I'm unsure how much I like the structure of the large for loop (iterating over `range(0, NUMBER_WEBSITES)` within the main command function. Looping over a list of urls in `_get_coupons` or `_get_coupons_from_page` might be a better alternative.

uqcsbot/dominos_coupons.py
tj-dodd: Good call

uqcsbot/dominos_coupons.py
tj-dodd: Not my proudest moment

uqcsbot/dominos_coupons.py
tj-dodd: Good call, I'll make it so they're comma separated - the bot originally had spaces separating the codes so I just assumed you wanted to stick with that style

49Indium: I had (incorrectly) assumed that the codes with spaces were just longer codes, not two different codes. Feel more than free to make stylistic changes as you see fit.

uqcsbot/dominos_coupons.py
tj-dodd: I really like this idea (I admit my implementation could be better), as well as the idea you left below in the unresolved comment. I'll begin working on this, and the code should be a lot cleaner afterwards.

tj-dodd:
An update:

- Requested changes implemented
- Now checks for and removes duplicate codes
- A good amount of restructuring in `__get_coupons_from_page`, `_get_coupons`, and `DominosCoupons`

uqcsbot/dominos_coupons.py
```diff
@@ -183,17 +158,62 @@ def keyword_matches(self, keyword: str) -> bool:
 
 
 def _get_coupons(
-    n: int, ignore_expiry: bool, keywords: List[str], source: str
-) -> List[Coupon]:
+    n: int,
+    ignore_expiry: bool,
+    keywords: List[str],
+    source: Literal["couponese", "frugalfeeds", "both"],
```
49Indium, line 164: Given that you use `Literal["couponese", "frugalfeeds", "both"]`, it might be best to create a new type (see [`advent.py`](https://github.com/UQComputingSociety/uqcsbot-discord/blob/main/uqcsbot/advent.py#L45) for an example).

uqcsbot/dominos_coupons.py
```diff
@@ -183,17 +158,62 @@ def keyword_matches(self, keyword: str) -> bool:
...
+            if isinstance(error, HTTPResponseException):
+                logging.warning(
+                    f"Received a HTTP response code {error.http_code}. Error information: {error}"
+                )
+                failed_urls.append(error.url)
 
     if not ignore_expiry:
         coupons = [coupon for coupon in coupons if coupon.is_valid()]
 
+    # Remove duplicates
```
49Indium, line 208: This is a great idea

uqcsbot/dominos_coupons.py
```diff
@@ -203,15 +223,18 @@ def _get_coupons(
...
 
-def _get_coupons_from_page(source: str) -> List[Coupon]:
+def _get_coupons_from_page(source: Literal["couponese", "frugalfeeds"]) -> List[Coupon]:
     """
     Strips results from html page and returns a list of Coupon(s)
     """
-    urls: List[str] = []
-    coupons: List[Coupon] = []
+
+    urls: Dict[str, str] = {
```
49Indium, line 234: I'd potentially make this dictionary a global constant, but that's up to you.

uqcsbot/dominos_coupons.py
```diff
@@ -226,69 +249,64 @@ def _get_coupons_from_page(source: str) -> List[Coupon]:
...
+        # Check if code is a digit - take seperators out so we can check
+        # this properly
+        temp_code: str = code.replace(",", "").replace(" ", "")
+        if not temp_code.isdigit():
+            continue
+
+        # Keep formatting same for coupons with multiple codes
+        if source == "frugalfeeds":
+            code = code.replace(",", ", ")
+        else:
```
49Indium, line 294: Instead of using `else`, it might be best to use `if source == "couponese"`, as this could be forgotten if we add more sources in the future.

uqcsbot/dominos_coupons.py
```diff
@@ -226,69 +249,64 @@ def _get_coupons_from_page(source: str) -> List[Coupon]:
...
+        code_container = soup_coupon.find(class_=siteclass.get("code"))
+
+        if not expiry_date_container or not description_container or not code_container:
+            continue
+
+        expiry_date_str: str = expiry_date_container.get_text(strip=True)
+        description: str = description_container.get_text(strip=True)
+        code: str = code_container.get_text(strip=True)
+
+        # Check if code is a digit - take seperators out so we can check
```
49Indium, line 285: Why do we need to check if the code is a digit? Is this for the pizza hut coupons (that FrugalFeeds sometimes provides)? If it is, state that explicitly in your comment.

49Indium (changes requested):
This is getting close to ready. I like the duplicate checking! A few small changes would be nice to make the typing more readable and reduce repetition.

uqcsbot/dominos_coupons.py
tj-dodd: Do you recommend a certain style? Is it fine to leave it as `urls` or should it be capitalized i.e. `URLS`? I'm aware constants should be fully capitalized but I'm unsure as it's a dictionary.

uqcsbot/dominos_coupons.py
tj-dodd: Yes, this is to deal with the pizza hut advertisements, I've since clarified this through a comment - FrugalFeeds have since taken down these advertisements but they might re-emerge in the future. They might also introduce some other form of advertising so I thought I should implement a somewhat robust filter, although there's probably a better way to do this so I'm open to any ideas.

uqcsbot/dominos_coupons.py
tj-dodd: I am fairly unfamiliar with creating new types, but I have had a crack at implementing this. I'm unsure if I'm following best practices here:

`CouponSource = Literal[
    "frugalfeeds", 
    "couponese", 
    "both"]`

I've declared this globally and have replaced all instances of `Literal["couponese", "frugalfeeds", "both"]` declaration with CouponSource, e.g. 

`source: CouponSource = "both"`, 
`def _get_coupons(
    n: int,
    ignore_expiry: bool,
    keywords: List[str],
    source: CouponSource,
) -> Tuple[List[Coupon], List[str]]:`,

And then have used `source` as I was using it before, e.g.. 

`coupons, failed_urls = _get_coupons(
            number_of_coupons, ignore_expiry, keywords.split(), source
        )`

`if source == "both":
            description_string = f"Sourced from [FrugalFeeds]({FRUGAL_FEEDS_DOMINOS_URL}) and [Couponese]({COUPONESE_DOMINOS_URL})"`

Am I on the right path here? Additionally, I spotted another possible area of improvement - I've used `Literal["couponese", "frugalfeeds"]]` a couple of times. Should I declare it as its own type as well? Is there a better way to approach this? It's just `CouponSource` without both.

tj-dodd:
Changes done 👍

uqcsbot/dominos_coupons.py
49Indium: I think this is good (especially with the new descriptive comment)

uqcsbot/dominos_coupons.py
49Indium: All capitalised would be my thoughts, but I'm not certain.

uqcsbot/dominos_coupons.py
49Indium: That's exactly it. It makes the code easier to read and reduces repetition. The `SingleSource` type is also a goo idea. Personally, I would have made `CouponSource = SingleSource | Literal["both"]`, but how you have implemented it is also very clean IMO.

49Indium (approved):

PR #211, opened 2024-11-20
Adjustments to Advent of Code

49Indium:
This consists of two parts (that could have been separate PRs but are related).

Firstly, better error are now given session token expires. If the session token expires, the leaderboard page redirects to the advent home page (with a successful 200 response code) so the bot fails to read this as python. If we block redirects, we can detect this earlier and provide a better error message.

Secondly, AOC registrations did not persist between years. By removing the year in the registrations table, registrations will persist between years. Not that this will also require access to the database to remove this column. PLEASE DON'T MERGE until this is done. Once reviewed, I am happy to do this.

### DIFF
--- a/uqcsbot/advent.py
+++ b/uqcsbot/advent.py
@@ -229,6 +229,7 @@ class Advent(commands.Cog):
             response = requests.get(
                 LEADERBOARD_URL.format(year=year, code=code),
                 cookies={"session": self.session_id},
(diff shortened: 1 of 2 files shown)


TheOnlyMrCat (dismissed):
Looks good to me.

andrewj-brown (approved):
ballin

PR #213, opened 2024-12-03
advent: increase interaction timeout and optimise

katrinafyi:
optimisation to cache image data when switching to/from full view. this is done by lru_cache, currently with a size of 16 (enough for 8 /leaderboard calls, with 2 image each)

### DIFF
--- a/uqcsbot/advent.py
+++ b/uqcsbot/advent.py
@@ -1,6 +1,6 @@
 import io
 import os
-from datetime import datetime
+from datetime import datetime, timedelta
 from random import choices
 from typing import Any, Callable, Dict, Iterable, List, Optional, Literal
 import requests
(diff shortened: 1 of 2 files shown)


katrinafyi:
i need help fixing this :pleading_face: 
```bash
/home/runner/work/uqcsbot-discord/uqcsbot-discord/uqcsbot/advent.py
  /home/runner/work/uqcsbot-discord/uqcsbot-discord/uqcsbot/advent.py:215:56 - error: Argument of type "Leaderboard" cannot be assigned to parameter "args" of type "Hashable" in function "__call__"
    "list[str | ColourFragment]" is incompatible with protocol "Hashable"
      "__hash__" is an incompatible type
      "__hash__" is not defined as a ClassVar in protocol
        Type "None" is incompatible with type "() -> int" (reportArgumentType)
1 error, 0 warnings, 0 informations 
```
i thought a dataclass would help it along but apparently not, maybe the type is too complex. pyright on my computer is wrongly flagging 5000+ errors in the stdlib..

TheOnlyMrCat (approved):
Lgtm :+1:

PR #214, opened 2024-12-04
advent: Add user-agent header to leaderboard requests

TheOnlyMrCat:
The AoC [automation guidelines](https://old.reddit.com/r/adventofcode/wiki/faqs/automation) specify that we should have a user-agent header on outgoing requests, so here is one.


JamesDearlove (approved):
Good catch, lgtm. We should have a user-agent most web requests like we do to UQ sites.

TheOnlyMrCat:
That's a good idea, I'll make a follow-up PR for that

PR #215, opened 2025-03-03
Do not load minecraft cog

PointerRain:
Currently, the minecraft cog is loaded every time the bot restarts, despite the Minecraft server not being a thing for years. This single character change will prevent it from being loaded on restarts. The cog can still be loaded and unloaded using the `/managecogs action:load cog:minecraft` command. If a new Minecraft server is created in the future, I encourage whoever is responsible to remove the offending `#`.


jenseni-git (approved):
lgtm

Issue #239, opened 2026-09-15
#yelling ban can be evaded through the creation of a poll.


bhavnoor0:
An exploit for the _#yelling_ channel was discovered by Amy where you can initiate a poll containing lowercase letters inside question and answers which is specifically forbid. [Source](https://discord.com/channels/813324385179271168/813412268191776798/1548972406872346635).

Reproducing issue
1. Go to [_#yelling_](https://discord.com/channels/813324385179271168/813412268191776798)
2. Create a new poll with lowercase characters in question and / or answers
3. The ban is evaded

PR #240, opened 2026-09-15
fix #yelling poll exploit

bhavnoor0:
This exploit was fixed by creating a new function which extracts all the text from poll question text and poll answers text if there is any and then combines all the text into a single string so that it can be cleaned by the clean_text function.
fixes issue #239.


uqcsbot/yelling.py
```diff
@@ -73,6 +73,19 @@ def __init__(self, bot: UQCSBot):
...
         )
 
+    def extract_text(self, msg: discord.Message) -> str:
+        """Extracts the input from text or poll from user message into text."""
+        extracted_text = []
+        if msg.content:
+            extracted_text.append(msg.content)
+
+        if msg.poll:
+            extracted_text.append(msg.poll.question.text)
```
PointerRain: `poll.question` is a str. It has no text attribute.

PointerRain (changes requested):
Good to see a fast fix to a recent issue, and great to see a new contributer.
Could you please also do a check for forwarded messages in your function. I believe they are called `message_snapshots` (for some reason unknown to me).

uqcsbot/yelling.py
bhavnoor0: Yes good catch.

bhavnoor0:
The pyright and formatting tests should be passing now. Sorry for not checking before hand. 
```fish
user@fedora ~/uqcsbot-discord (main)> poetry run black uqcsbot
All done! ✨ 🍰 ✨
42 files left unchanged.
user@fedora ~/uqcsbot-discord (main)> poetry run pyright uqcsbot
WARNING: there is a new pyright version available (v1.1.409 -> v1.1.414).
Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`

0 errors, 0 warnings, 0 informations
```

PointerRain (approved):
Good work.

